PR #51, opened 2020-01-14
Feature/direct request link

sampsapenna:
### Description

Added support for linking to the request creation directly, to enable the creation of shared access requests with reduced user interaction/input.

### Type of change

### Changes Made

* Implement login re-direction support
* Add a landing page for the direct shared access requests
* Fix footer location on sharing functionality related views
* Improve translations
* Add a dockerfile for development builds (building the UI without minifying and optimization)

### Testing

- [x] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply

### Mentions

@blankdots – Debugging footer css

### DIFF
--- a/swift_browser_ui_frontend/src/views/DirectRequest.vue
+++ b/swift_browser_ui_frontend/src/views/DirectRequest.vue
@@ -2,6 +2,7 @@
   <section>
     <b-field
       v-if="projects.length > 1"
+      horizontal
       label="Project"
     >
       <b-select
@@ -13,7 +14,7 @@
           :key="item.id"
           :value="item.id"
         >
-          {{ project.name }}
+          {{ item.name }}
         </option>
       </b-select>
     </b-field>
@@ -43,7 +44,9 @@
       />
     </b-field>
 
-    <b-field>
+    <b-field
+      horizontal
+    >
       <p class="control">
         <button
           class="button is-primary"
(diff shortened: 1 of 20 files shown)


swift_browser_ui_frontend/src/views/DirectRequest.vue
```diff
@@ -0,0 +1,133 @@
...
+      } catch (ReferenceError) {
+        delay(this.setParams, wait, [wait * 2]);
+      }
+    },
+    checkMultiProject: function (wait) {
+      try {
+        if (this.projects.length > 1) {
+          this.$buefy.notification.open({
+            indefinite: true,
+            message: "Account has access to multiple projects. " +
```
blankdots: maybe this could also be in `lang` part ?

swift_browser_ui_frontend/src/views/ShareRequests.vue
```diff
@@ -1,13 +1,13 @@
 <template>
-  <section>
+  <div class="contents">
```
blankdots: `section` would have worked as well because it is not related to flow but more towards sectioning content, see:  https://html.spec.whatwg.org/multipage/indices.html#element-content-categories

swift_browser_ui_frontend/src/views/DirectRequest.vue
sampsapenna: True, I'll change that over to the translations.

swift_browser_ui_frontend/src/views/ShareRequests.vue
sampsapenna: Yup, seems to be so as I read more into it, I'll revert back to section as it's more descriptive

PR #79, opened 2020-05-22
Merge scoped API token support

sampsapenna:
### Description

PR adds scoped API tokens, to enable the creation of `sharing` and `shared-access-request` tokens in the UI, for use in the sharing tools implemented in [swift-sharing-tools](https://github.com/cscfi/swift-sharing-tools). PR also contains contents of PR #77 (pseudofolder rendering) and PR #78 (unit tests for `signature.py` module)

### Type of change

### Changes Made

* Add support for creating scoped API tokens in the UI
    - Scoped API tokens can be added by navigating to the user information dashboard, and clicking the `Tokens` button
    - The tokens can be removed from the APIs, but once created the original token can't be retrieved
* #77 add pseudofolder rendering
    - Pseudofolder rendering also fixes the issue of page changing with page number changing in route
* Unit tests for the `signature.py` module

### Testing

- [x] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply


swift_browser_ui/signature.py
```diff
@@ -29,12 +37,168 @@ async def handle_signature_request(
...
+) -> aiohttp.web.Response:
+    """Handle call for an API token create."""
+    session = api_check(request)
+
+    project = request.app["Creds"][session]["active_project"]["id"]
+
+    LOGGER.debug(f"Creating a scoped API token for {project}")
+
+    ident = request.match_info["id"]
+    token = hashlib.sha256(os.urandom(256)).hexdigest()  # nosec
```
blankdots, line 51: I think this https://docs.python.org/3.7/library/secrets.html would be more useful here.

swift_browser_ui/server.py
```diff
@@ -157,6 +162,13 @@ async def servinit() -> aiohttp.web.Application:
         aiohttp.web.get('/sign/{valid}', handle_signature_request)
     ])
 
+    # Add token functionality
+    app.add_routes([
```
blankdots, line 166: Question: In order to use these endpoints, all it requires is for a login to be valid or also for additional permissions ?

swift_browser_ui_frontend/src/components/ObjectTable.vue
```diff
@@ -0,0 +1,560 @@
...
+                  size="is-small"
+                /> {{ $t('message.downloadLink') }}
+              </a>
+            </li>
+          </ul>
+        </span>
+      </template>
+      <template slot="empty">
+        <p
+          style="width:100%;text-align:center;margin-top:5%;margin-bottom:5%"
```
blankdots, line 256: can this be handled by an CSS id ?

swift_browser_ui_frontend/src/components/ObjectTable.vue
```diff
@@ -0,0 +1,560 @@
...
+        </p>
+      </div>
+    </b-field>
+    <b-table
+      focusable
+      detailed
+      hoverable
+      narrowed
+      header-checkable
+      style="width: 90%;margin-left: 5%; margin-right: 5%;"
```
blankdots, line 68: should/can this be handled within a CSS class or id ?

swift_browser_ui_frontend/src/views/Tokens.vue
```diff
@@ -0,0 +1,143 @@
...
+            >
+              {{ $t('message.tokens.createToken') }}
+            </button>
+          </p>
+        </b-field>
+      </b-field>
+    </div>
+    <div
+      v-if="latest"
+      style="width: 90%;margin-left: 5%; margin-right: 5%;"
```
blankdots, line 42: CSS id ?

swift_browser_ui/server.py
sampsapenna: Yup, all that's required is a valid login – the tokens created are project scoped, so access to said project should be enough

PR #90, opened 2020-07-08
bump to 1.0.0rc3

blankdots:
### Type of change

### Changes Made

see #86 #87 #88 #89 

### Testing

- [x] Unit Tests


blankdots:
tested in staging.

.github/workflows/eslint.yml
```diff
@@ -0,0 +1,27 @@
...
+
+on: [push]
+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [ '12', '14' ]
```
otahontas, line 11: we do we run linting for both versions?

blankdots: it was designed and built for `12` (https://github.com/CSCfi/swift-browser-ui/blob/devel/Dockerfile#L1) but I want to see if we can easily move to newer versions such as `14` without breaking anything.

At some point will move directly to `14`, till then will be testing both.

Issue #117, opened 2020-12-01
catch swiftclient.exceptions.ClientException


blankdots:
#### Describe the bug

User is shown not authorized if it goes to project not authorized for swift

#### To Reproduce
Steps to reproduce the behaviour:
1. Go to 'project not authorized for openstack swift'
```
ui_1        | [2020-12-01 09:02:54][swiftclient.service][10 MainProcess][ERROR   ] (L:976) _list_account_job: Account GET failed: https://<url>?format=json 401 Unauthorized  [first 60 chars of response] b'{"Code":"AccessDenied","RequestId":"<redacted>'
ui_1        | Traceback (most recent call last):
ui_1        |   File "/usr/local/lib/python3.7/site-packages/swiftclient/service.py", line 951, in _list_account_job
ui_1        |     headers=req_headers
ui_1        |   File "/usr/local/lib/python3.7/site-packages/swiftclient/client.py", line 1859, in get_account
ui_1        |     delimiter=delimiter)
ui_1        |   File "/usr/local/lib/python3.7/site-packages/swiftclient/client.py", line 1802, in _retry
ui_1        |     service_token=self.service_token, **kwargs)
ui_1        |   File "/usr/local/lib/python3.7/site-packages/swiftclient/client.py", line 882, in get_account
ui_1        |     raise ClientException.from_response(resp, 'Account GET failed', body)
ui_1        | swiftclient.exceptions.ClientException: Account GET failed: https://<url>?format=json 401 Unauthorized  [first 60 chars of response] b'{"Code":"AccessDenied","RequestId":"<redacted>'
```
#### Expected behaviour

Show a message that user is not authorized for this project.

Catch exception in other methods as well

blankdots:
difficult to catch that exception, a better idea would be to address the list result `i["success"]` and `i["error"]` in a different function however it does not provide much value, as if that project is enabled for the user but there is no access, containers will still not be accessible.

A better idea will be to filter the project if the user does not have access, but it is also difficult to achieve.

Resuming to extending the empty message to reflect that the project might not be enabled for user to use containers.

blankdots:
closing as won't fix for now as it is not a breaking error

PR #120, opened 2021-01-25
bump to 1.0.0.rc12

blankdots:
### Type of change

### Changes Made

1. update npm packages
2. catch more exceptions
3. use python 3.8 in docker image
4. attempted to address #117 at least by a message

### Testing

- [x] Tests do not apply

### Mentions

exception pointed out in #117 is difficult to catch, a better idea would be to address the list result `i["success"]` and `i["error"]` in a different function, however it does not provide much value (besides being Pythonic and we can treat the exception at a later point in time), as if that project is enabled for the user but there is no access, containers will still not be accessible.


swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -309,7 +310,8 @@ const translations = {
                         "nykyisen kirjautumisen ajaksi.",
       largeDownAction: "Hyväksy",
       emptyContainer: "Säiliö on tyhjä.",
-      emptyProject: "Projektilla ei ole säiliöitä.",
+      emptyProject: "Projektilla ei ole säiliöitä " +
+                    "tai niitä ei ole sallittu tankkien käyttämiseen.",
```
teemukataja, line 314:
```suggestion
                    "tai niiden käyttöä ei ole sallittu.",
```

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -134,7 +134,8 @@ const translations = {
                         "enable them for the duration of the session.",
       largeDownAction: "Enable",
       emptyContainer: "This container is empty.",
-      emptyProject: "The project does not contain any containers.",
+      emptyProject: "The project does not contain any containers " +
+                    "or not enabled for using containers.",
```
teemukataja, line 138:
```suggestion
                    "or their use is not permitted.",
```

teemukataja (approved):

teemukataja (approved):

PR #124, opened 2021-02-08
bump to 1.0.0rc15

blankdots:
### Description

update docker image with rust in build, needed by cryptography package https://cryptography.io/en/latest/installation.html#rust
There was an `OSError` exception in health endpoint when other endpoints are down so decided to catch them all (shameless Pokemon reference)

### Type of change

### Changes Made

1. breaking changes in cryptography require rust > 1.45 which is in alpine 3.13
2. updated npm packages
3. catch all exceptions in health call

### Testing

- [x] Tests do not apply

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,17 +19,14 @@ RUN cd /root/swift_ui/swift_browser_ui_frontend \
 FROM python:3.8-alpine3.13 as BACKEND
 
 RUN apk add --update \
-    && apk add --no-cache build-base curl-dev linux-headers bash git\
-    && apk add --no-cache libressl-dev libffi-dev\
+    && apk add --no-cache build-base curl-dev linux-headers bash git \
+    && apk add --no-cache libressl-dev libffi-dev rust cargo \
     && rm -rf /var/cache/apk/*
 
 COPY requirements.txt /root/swift_ui/requirements.txt
 COPY setup.py /root/swift_ui/setup.py
 COPY swift_browser_ui /root/swift_ui/swift_browser_ui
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -28,11 +28,14 @@ COPY setup.py /root/swift_ui/setup.py
...
+    && rm -rf /var/cache/apk/*
+
 RUN pip install --upgrade pip && \
     pip install -r /root/swift_ui/requirements.txt && \
     pip install /root/swift_ui
 
-FROM python:3.8-alpine3.12
+FROM python:3.8-alpine3.13
 
 RUN apk add --no-cache --update bash
```
vvaltia, line 41: Not required for right now, but would make more sense to have just one docker image. If the files differ, they could be added as configmaps instead.

swift_browser_ui/health.py
```diff
@@ -120,6 +132,12 @@ async def get_upload_runner(
         services["swiftui-upload-runner"] = {
             "status": "Error"
         }
+    except Exception as e:
+        request.app["Log"].debug("Poll swift-x-account-sharing failed")
+        request.app["Log"].info(f"reason: {e}")
+        services["swift-sharing-request"] = {
+            "status": "Error"
+        }
```
vvaltia, line 140: Imo this would make a good function, since the exact same exception is here multiple times

vvaltia (dismissed):

swift_browser_ui/health.py
blankdots: fair enough will add that.

teemukataja (dismissed):

Dockerfile
blankdots: for this dockerfile `BACKEND` builds the packages while the layer you pointed has the binaries
If we take into consideration other swift-ui services sure, that could be a way.

PR #167, opened 2021-04-15
Bugfix/fix object refresh after upload

### LINKED ISSUE #110: No list refresh after upload
When uploading a file to an bucket (Firefox + Windows), it gives a message that transfer was OK, but **the file does not appear unless one refreshes the folder from the browser.**  It does not even help to go to another folder meanwhile.

In my opinion this is a showstopper to release this for public use.

sampsapenna:
### Description

Fixes issue #110 – the objects in a container are now dynamically refreshed whenever an object upload completes. The refreshes are done upon completion of each uploaded file. File refreshes work also inside pseudofolders.
Additionally the PR fixes a newfound issue in the upload button, which did not re-register to the UI correctly after being replaced with upload cancellation, and implements uploads into pseudofolders from the button, and from drag-n-drop.

### Related issues

* Fixes #110 

### Type of change

### Changes Made

* Move object listing to Vuex to make allow updating objects outside `ObjectTable` component
* Remove caching object listings since checking for new content requires a full `GET` with current backend implementation, making caching somewhat unnecessary
* Move upload button from `FolderUploadForm` into its own component, to enable re-registration to `Resumable` object upon reactivation in the UI when an upload completes
* Rewrite `getUploadUrl` in `main.js` to enable uploading single files into pseudofolders (+ folders with drag-n-drop)
* Removed some unnecessary stuff

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/common/store.js
+++ b/swift_browser_ui_frontend/src/common/store.js
@@ -2,7 +2,7 @@
 import Vue from "vue";
 import Vuex from "vuex";
 
-import { recursivePruneCache } from "@/common/conv";
+// import { recursivePruneCache } from "@/common/conv";
 import { getBuckets } from "@/common/api";
 import { 
   getObjects,
@@ -19,8 +19,7 @@ const store = new Vuex.Store({
     multipleProjects: false,
     isLoading: false,
     isFullPage: true,
-    objectCache: {},
-    sharedObjects: undefined,
+    objectCache: [],
     containerCache: [],
     langs: [
       {ph: "In English", value: "en"},
@@ -33,15 +32,6 @@ const store = new Vuex.Store({
     isChunking: false,
     uploadProgress: undefined,
     altContainer: undefined,
-  },
-  getters: {
-    getObjectsByContainer: (state) => (container) => {
-      return (
-        state.objectCache[container] != undefined ?
-          state.objectCache[container] :
-          []
-      );
-    },
   },
   mutations: {
     updateContainers (state) {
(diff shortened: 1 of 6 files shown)


swift_browser_ui_frontend/src/common/store.js
```diff
@@ -2,8 +2,12 @@
 import Vue from "vue";
 import Vuex from "vuex";
 
-import { recursivePruneCache } from "@/common/conv";
+// import { recursivePruneCache } from "@/common/conv";
```
blankdots, line 5: is this a forgotten comment?

swift_browser_ui_frontend/src/common/store.js
```diff
@@ -43,16 +47,50 @@ const store = new Vuex.Store({
...
+        ).catch(() => {
+          state.isLoading = false;
+        });
+      }
+      else {
+        getObjects(container).then((ret) => {
+          if (ret.status != 200) {
+            state.isLoading = false;
+          }
+          // state.objectCache = recursivePruneCache(state.objectCache);
```
blankdots, line 84: same here ?

swift_browser_ui_frontend/src/common/store.js
sampsapenna: Yup, from when I was thinking if I should re-implement the caching anyway

I'll remove those

sampsapenna: I think that this will make `recursivePruneCache` unused, so that should probably go as well

sampsapenna: Removed 👍🏼

swift_browser_ui_frontend/src/common/store.js
sampsapenna: Yup, removed 👍🏼

blankdots (approved):

Issue #111, opened 2020-11-03
Redirect from login if user session is active


blankdots:
If user session is active don't show login screen, instead forward to `/browse` directly.

sampsapenna:
Implemented in PR #171

PR #171, opened 2021-04-23
Feature/redirect existing sessions

sampsapenna:
### Description

Valid logged-in sessions were previously non-persistent. I.e. people with open sessions were forced to log in again instead of being redirected to the UI. This behavior has now been fixed, and established sessions are persistent with a inaction period of 1h and a total session lifetime of 8h, matching that of Openstack.

### Related issues

* Implements / fixes issue #111 

### Type of change

### Changes Made

* Make sessions persistent during lifetime
* Refactor session invalidation into a function
* Check for session lifetime instead of trusting Openstack to invalidate in time
* Remove separate set for session ids as unnecessary
* Refactor tests to account for session redirection
* Reduce token length from 512 bits to 256 as previous was unnecessarily long and the shorter one is secure enough
* Change tox *flake8* rules in preparation for moving to *black* coding style

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui/login.py
+++ b/swift_browser_ui/login.py
@@ -13,10 +13,17 @@ from multidict import MultiDictProxy
 import typing
 import urllib.error
 
-from ._convenience import disable_cache, decrypt_cookie, generate_cookie
-from ._convenience import get_availability_from_token, session_check
-from ._convenience import initiate_os_session, initiate_os_service
-from ._convenience import test_swift_endpoint
+from ._convenience import (
+    disable_cache,
+    decrypt_cookie,
+    generate_cookie,
+    get_availability_from_token,
+    session_check,
+    initiate_os_session,
+    initiate_os_service,
+    test_swift_endpoint,
+    clear_session_info,
+)
 from .settings import setd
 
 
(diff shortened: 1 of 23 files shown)


swift_browser_ui/login.py
```diff
@@ -165,69 +173,79 @@ async def sso_query_end(
...
-    request.app['Creds'][session]['active_project'] = {
+    request.app['Sessions'][session]['active_project'] = {
         "name": project_name,
         "id": project_id
     }
 
+    # Save time of login to the backend
+    created = time.time()
+    request.app["Sessions"][session]["last_used"] = created
+    request.app["Sessions"][session]["max_lifetime"] = created + 28800
```
blankdots, line 247: `28800` is a value that we can keep in `setd` in `settings.py` and be reused here and in `set_cookie` `max_age`

swift_browser_ui/login.py
```diff
@@ -303,7 +321,7 @@ async def token_rescope(
             response.headers["Location"] = request.headers["Referer"]
     response.set_cookie(
         "LAST_ACTIVE",
-        request.app['Creds'][session]['active_project']['id'],
+        request.app['Sessions'][session]['active_project']['id'],
         expires=2592000  # type: ignore
```
blankdots, line 325: same could be in `settings.py` `setd` as it is in multiple places

blankdots (commented):
small comments, remains to build and test to see how it works.

blankdots (approved):

PR #241, opened 2021-06-15
change naming from containers to buckets

sampsapenna:
### Description

For most users the term `bucket` for a container is more easily understood.

### Type of change

### Changes Made

* Change naming convention from `container` -> `bucket`

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply


blankdots (approved):
good. however are we fine with `objects` name or do we use something else?

sampsapenna:
> good. however are we fine with `objects` name or do we use something else?

`objects` is fine, as that's used with all object storage solutions already

PR #267, opened 2021-07-07
Feature/support credential login

sampsapenna:
### Description

PR implements support for configurable credential based login to enable logging in with conventional username and password, adding this as a required login method besides the configured federated login methods in keystone. Credentials are used in a similar manner to how keystone library operates, i.e. username and password are only used for fetching an unscoped token from the authentication backend and after this used similarly to token logins.

### Type of change

### Changes Made

* Add backend support for username+password logins
* Add support for logging in with username in the dev login page
* Add a properly styled page for logging in with username
* Update unit and ui tests to test for the new login method
* Update language customizations to deal with the new login method

### Testing

- [x] Unit Tests
- [x] Integration Tests
- [ ] Tests do not apply

### DIFF
--- a/swift_browser_ui/_convenience.py
+++ b/swift_browser_ui/_convenience.py
@@ -338,7 +338,7 @@ def os_get_token_from_credentials(
         setd["auth_endpoint_url"],
         username=username,
         password=password,
-        user_domain_name="Default",
+        user_domain_name=setd["os_user_domain"],
         unscoped=True,
     )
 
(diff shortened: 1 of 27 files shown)


swift_browser_ui/_convenience.py
```diff
@@ -329,6 +329,27 @@ def initiate_os_session(unscoped: str, project: str) -> keystoneauth1.session.Se
...
+def os_get_token_from_credentials(
+    username: str,
+    password: str,
+) -> str:
+    """Get an unscoped token with provided credentials."""
+    os_auth = v3.Password(
+        setd["auth_endpoint_url"],
+        username=username,
+        password=password,
+        user_domain_name="Default",
```
blankdots, line 341: this needs to be configurabe as i think this openstack installation specific

sampsapenna: has been made configurable in the lates commit

blankdots (approved):
need to catch empty user and password but other than that approving preemptively

PR #394, opened 2021-11-12
Return empty list of buckets, fixes #356

### LINKED ISSUE #356: Deleting the last bucket does not refresh the browser view
#### Describe the bug
If you delete the last bucket in the project a notification is shown that the bucket is deleted, but the view does not get updated and the bucket is still visible and you can still open the bucket.

#### To Reproduce
1. Select a project with one bucket (or one without any and create a new bucket).
2. Select the bucket and click "Delete".
3. A notification pops up saying bucket was deleted.
4. The bucket is still visible the browser and is still accessible, you can click "Delete" again and you can even open the bucket.

#### Expected behavior
Along with the notification in step 3 above, the browser view should be refreshed so that the bucket disappears.

#### Screenshots
image

#### Additional context
Tested with latest Firefox (93.0 for macOS).

csc-felipe:
### Description

The backend was returning an HTTP 404 when there were no buckets.
This caused the frontend to not refresh the bucket list UI when the last bucket was deleted.

### Related issues
Fixes #356.

### Type of change

### Changes Made

Removed HTTP 404 return when there are no buckets.

### Testing

- [x] Unit Tests

### DIFF
--- a/swift_browser_ui/ui/api.py
+++ b/swift_browser_ui/ui/api.py
@@ -67,11 +67,6 @@ async def swift_list_buckets(request: aiohttp.web.Request) -> aiohttp.web.Respon
         # response at once
         serv = request.app["Sessions"][session]["ST_conn"].list()
         [_unpack(i, cont, request) for i in serv]
-        # for a bucket with no objects
-        if not cont:
-            # return empty object
-            request.app["Log"].debug("Empty container list.")
-            raise aiohttp.web.HTTPNotFound()
         return aiohttp.web.json_response(cont)
     except SwiftError:
         request.app["Log"].error("SwiftError occured return empty container list.")
--- a/tests/ui_unit/test_api.py
+++ b/tests/ui_unit/test_api.py
@@ -107,10 +107,11 @@ class APITestClass(asynctest.TestCase):
         )
 
     async def test_list_without_containers(self):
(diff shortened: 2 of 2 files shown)


blankdots:
I would like to see a cypress tests on this ... so we make sure it does not happen again. see https://github.com/CSCfi/swift-browser-ui/tree/master/tests/cypress/integration and [cy](https://www.cypress.io/) for docs

csc-felipe:
Even though the symptom occurred in the frontend (table view didn't refresh). The cause was that the backend was sending something (HTTP 404) different than what the frontend expected (empty list). And this is covered by the unit test in the PR. 

To create cypress tests for this, I would need to add a `delete` method to the `Mock_server()`, so it wouldn't use the actual code which could cause the real issue. So even if we would come across a similar issue in the future, the cypress tests would never catch it, because it runs the mock server, and not the real one. I'm thinking that it might not make sense to implement this. What do you think?

Furthermore, adding tests for this would be more about testing that the deleting functionality in the UI works than that the list is empty. So it would be a good task when adding more e2e tests.

blankdots:
@csc-felipe 
>  I'm thinking that it might not make sense to implement this. What do you think?

make sense, I was asking if it is possible

> So it would be a good task when adding more e2e tests.
please add a task about it

blankdots (approved):

PR #418, opened 2021-11-30
Fix breadcrumbs by changing URLs

### LINKED ISSUE #417: Breadcrumbs are broken in some pages
#### Describe the bug
Currently, the URLs are not consistent with each other, which breaks the breadcrumbs creation and links. The breadcrumbs builder function currently requires user, project and when is the case, container and object in the URL. 

#### To Reproduce
1. Go to different pages and look at the breadcrumbs and the links under them.

#### Expected behavior
All the pages have similar URLs, and breadcrumbs are consistent across pages.

csc-felipe:
### Description
Making URLs look similar, basically adding parameters to the ones that didn't have them.
This is a relatively simple change which makes the frontend pass on more parameters between pages. The breadcrumb function expects these parameters to be available. When they are not, the functionality is broken.

### Related issues
Fixes #417.

### Type of change

### Changes Made

Changed the routes in the frontend to include more parameters.

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/common/router.js
+++ b/swift_browser_ui_frontend/src/common/router.js
@@ -40,7 +40,7 @@ export default new Router({
       component: SharingView,
     },
     {
-      path: "/browse/:user/:project/:container/add",
+      path: "/browse/:user/:project/add",
       name: "AddContainer",
       component: CreateContainer,
     },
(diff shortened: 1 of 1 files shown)


blankdots (commented):
i am ok with the change, however not sure how relevant is the user in this case as most of the buckets and containers are project specific not user, user is just a way to access the project and identify whom did what :thinking:

blankdots (approved):

Issue #85, opened 2020-06-18
Using metadata to organize container files


sampsapenna:
#### Proposed solution

The metadata on the swift objects and containers can be used to include information in the objects on how to organize them, e.g. flagging important and log files. This information can then be used e.g. for better search functionality (in the future).

As the backend already supports querying the metadata, this should be purely a frontend task  (though API handlers for adding / altering the metadata might need to be added)

#### DoD (Definition of Done)

The objects can include some additional categorization etc. which can be displayed e.g. as tags under the details tab in the table. 

#### Testing

* Visual inspection of the implemented functionality.

PR #419, opened 2021-11-30
Allow a person to tag buckets.

csc-felipe:
### Description
Gives a person the possibility to add or remove tags from a bucket. Includes the ability to show or hide tags when browsing the buckets.

Please, check the Finnish language :) 

### Related issues
Related to #85.

### Type of change

### Changes Made
Added backend for updating metadata for a bucket. Added UI for displaying tags for containers, as well as hiding them. There is a new button, called 'Edit', which makes it possible to edit a bucket's tags. When creating a new Bucket, it's possible to add tags to it.

Tags are not displayed in the bucket view (after double click on a bucket). There's no edit button on that page either.

### Testing

- [x] Unit Tests
- [x] Integration Tests

### Mentions
Thanks to @sampsapenna for solving my javascript sync/async mess.

### DIFF
--- a/swift_browser_ui_frontend/src/common/lang.js
+++ b/swift_browser_ui_frontend/src/common/lang.js
@@ -63,6 +63,7 @@ let default_translations = {
         created: "Created",
         folderDetails: "No details for folders",
         clearChecked: "Clear checked",
+        showTags: "Display tags",
       },
       discover: {
         sync_shares: "Synchronize shared buckets",
@@ -185,12 +186,15 @@ let default_translations = {
       copy: " Copy",
       create: "Create",
       delete: "Delete",
+      edit: "Edit",
+      save: "Save",
       createContainerButton: "Create Bucket",
       copysuccess: "Started copying the bucket in the background",
       copyfail: "Failed to copy the bucket",
       renderFolders: "Render as Folders",
       container_ops: {
         addContainer: "Add a new bucket",
+        editContainer: "Editing bucket: ",
         deleteConfirm: "Delete Bucket",
         deleteConfirmMessage: "Are you sure you want to delete this " +
                               "bucket?",
@@ -199,6 +203,8 @@ let default_translations = {
         containerMessage: "The name of the new bucket",
         fullDelete: "Deleting a bucket with contents requires deleting " +
                     "all objects inside it first.",
+        tagName: "Tags",
+        tagMessage: "Press enter to add.",
       },
       objects: {
         deleteConfirm: "Delete Objects",
@@ -286,6 +292,7 @@ let default_translations = {
         created: "Luotu",
         folderDetails: "Ei yksityiskohtia kansioille",
         clearChecked: "Poista valinnat",
+        showTags: "Näytä Tägit",
       },
       discover: {
         sync_shares: "Synkronoi jaetut säiliöt",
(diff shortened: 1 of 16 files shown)


csc-felipe:
cypress is clicking the wrong switch 
Browse buckets and test operations -- should be able to filter table, adjust display buckets per page and pagination (failed)
.

blankdots:
> cypress is clicking the wrong switch

then we need a better selector for that switch, or something is off with clicking the parent `cy.contains('Paginated').parent().click()` because you put them in the same `control` div, moving them to separate divs should help or finding out a new selector

swift_browser_ui_frontend/src/views/Containers.vue
```diff
@@ -33,12 +33,17 @@
         </option>
       </b-select>
       <div class="control is-flex">
-        <b-switch 
+        <b-switch
```
blankdots, line 36: if you want tests to pass this should help:
```
<div class="control is-flex">
        <b-switch
          v-if="bList.length < 500"
          v-model="isPaginated"
        >
          {{ $t('message.table.paginated') }}
        </b-switch>
      </div>
      <div class="control is-flex">
        <b-switch
          v-model="showTags"
        >
          {{ $t('message.table.showTags') }}
        </b-switch>
      </div>
```

but switches can be grouped also as: https://buefy.org/documentation/switch#styles 

however maybe we need to select that switch a tad different than `cy.contains('Paginated').parent().click()` ... maybe add a `data-testid=` attribute to it

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -409,18 +416,23 @@ let default_translations = {
       copy: " Kopioi",
       create: "Luo",
       delete: "Poista",
+      edit: "Muokkaa",
+      save: "Tallena",
```
sampsapenna, line 420: ~Tallena~ -> Tallenna

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -286,6 +292,7 @@ let default_translations = {
         created: "Luotu",
         folderDetails: "Ei yksityiskohtia kansioille",
         clearChecked: "Poista valinnat",
+        showTags: "Näytä tagit",
```
sampsapenna, line 295: Probably better as "tägit" with the umlaut – the anglicism is a bit informal/colloquial, but is probably more recognizable for the user

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -409,18 +416,23 @@ let default_translations = {
...
       delete: "Poista",
+      edit: "Muokkaa",
+      save: "Tallena",
       createContainerButton: "Luo säiliö",
       copysuccess: "Aloitettiin säiliön kopiointi taustalla",
       copyfail: "Säiliön kopiointi epäonnistui",
       renderFolders: "Näytä kansioina",
       container_ops: {
         addContainer: "Luo uusi säiliö",
+        editContainer: "Muokkaaminen säiliö: ",
```
sampsapenna, line 427: *Muokataan säiliötä:

sampsapenna (changes requested):
Some small alterations to the Finnish upon further inspection, other than that LGTM.

swift_browser_ui_frontend/src/common/lang.js
csc-felipe: Also on the field label that says Tagit, or only here?

swift_browser_ui_frontend/src/views/AddContainer.vue
```diff
@@ -16,6 +20,24 @@
...
+    >
+      <b-taginput
+        v-model="tags"
+        ellipsis
+        maxlength="20"
+        icon="label"
+        has-counter
+        rounded
+        type="is-primary"
+        :confirm-keys="[',', ';', ':', '.', ' ', 'Tab', 'Enter']"
```
blankdots, line 39: can we create a var out of these keys so we can manage them a tad better ?

blankdots (changes requested):
the new POST endpoint "/api/bucket/meta" should be added to the [API.yaml](https://github.com/CSCfi/swift-browser-ui/blob/master/docs/_static/api.yml)

swift_browser_ui_frontend/src/common/lang.js
sampsapenna: Preferably change that to "Tägit" as well

blankdots (approved):

sampsapenna (approved):

PR #421, opened 2021-12-03
Refactor: pass icon as props to button.

csc-felipe:
### Description

Pass icons as props to buttons that didn't yet, so that they position better in relation to the button label. I think it looks better, but it does take up more horizontal space.

Before:
image
After:
image

### Type of change

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/components/ContainerDownloadLink.vue
+++ b/swift_browser_ui_frontend/src/components/ContainerDownloadLink.vue
@@ -7,11 +7,9 @@
     target="_blank"
     :inverted="inverted"
     disabled
+    icon-left="download"
   >
-    <b-icon
-      icon="download"
-      size="is-small"
-    /> {{ $t('message.downloadContainer') }}
+    {{ $t('message.downloadContainer') }}
   </b-button>
   <b-button
     v-else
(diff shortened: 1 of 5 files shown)


sampsapenna (approved):

blankdots (changes requested):
I think we missed a few spots e.g. in ObjectTable lines 257, 268, 278

search in frontend dir `icon=` and see what else can be improved

csc-felipe:
> I think we missed a few spots e.g. in ObjectTable lines 257, 268, 278
> 
> search in frontend dir `icon=` and see what else can be improved

@blankdots I had already checked those. They are HTML <a> tags, not buefy buttons. Thus the change does not apply.

Or was your suggestions to convert them to buttons?

blankdots:
> > I think we missed a few spots e.g. in ObjectTable lines 257, 268, 278
> > 
> > search in frontend dir `icon=` and see what else can be improved
> 
> @blankdots I had already checked those. They are HTML `<a>` tags, not buefy buttons. Thus the change does not apply.
> 
> Or was your suggestions to convert them to buttons?

it is fine if they are as <a>

blankdots (approved):

Issue #69, opened 2020-02-27
New shared access / access request alerts


sampsapenna:
#### Proposed solution

The user would greatly benefit from some kind of a notification (e.g. toast) whenever new sharing and request API related information is available.

#### DoD (Definition of Done)

The user is reminded of new sharing and request API related actions upon logging in.

#### Testing

Manual testing of UI behavior.

PR #422, opened 2021-12-03
Automatically refresh ACL information in sharing db

### LINKED ISSUE #305: Better ACL information synchronization
#### Proposed solution

ACL information doesn't currently synchronize automatically, this could be improved by having the sync happen automatically in the background. Container deletion doesn't synchronize with the sharing API, when deleting a container in the UI – this functionality needs to be added.

#### DoD (Definition of Done)

Shared container information is synchronized automatically in the background, including container deletions.

#### Testing

Can be quickly verified by hand.

sampsapenna:
### Description

Add automatic synchronization for ACL information in sharing database. I.e. current container access additions, removals, edits and container removals.

Previously the sharing information had a tendency towards containing stale data, since people would also access the sharing information in the CLI, remove containers, etc. leading to ACLs changing without updates.

### Related issues

* Fixes #305 

### Type of change

### Changes Made

* Make ACL synchronization run automatically upon login (scheduled on main App creation event)
* Make ACL synchronization a proper background operation to prevent unresponsive UI during sync (force yielding execution in container ACL metadata fetch loop)

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply
- [x] Needs testing (start an issue or do a follow up PR about it) #424

### DIFF
--- a/swift_browser_ui_frontend/src/common/conv.js
+++ b/swift_browser_ui_frontend/src/common/conv.js
@@ -64,15 +64,10 @@ export async function syncContainerACLs(client, project) {
   // Delete stale shared container access entries from the database
   for (let container of currentsharing) {
     if (!(Object.keys(aclmeta).includes(container))) {
-      let resp = await client.shareContainerDeleteAccess(
+      await client.shareContainerDeleteAccess(
         project,
         container,
       );
-      if (!resp) {
-        console.log("Container access deletion failed");
-      }
-    } else {
-      continue;
     }
   }
 
--- a/swift_browser_ui_frontend/src/common/conv.js
+++ b/swift_browser_ui_frontend/src/common/conv.js
@@ -42,8 +42,31 @@ export async function syncContainerACLs(client, project) {
   let aclmeta = acl.access;
   let currentsharing = await client.getShare(project);
 
+  // Delete stale shared access entries from the database
+  console.log(currentsharing);
+  for (let container of currentsharing) {
+    console.log("checking container existence for ", container);
+    console.log(aclmeta);
+    console.log("compare: ", !(Object.keys(aclmeta).includes(container)));
+    if (!(Object.keys(aclmeta).includes(container))) {
+      console.log("container", container, " no longer exists, removing");
+      let resp = await client.shareContainerDeleteAccess(
+        project,
+        container,
+      );
+      if (resp) {
+        console.log("container access deletion successful");
+      } else {
+        console.log("container access deletion failed");
+      }
+    } else {
+      continue;
+    }
+  }
+
   // sync new shares into the sharing database
   for (let container of Object.keys(aclmeta)) {
+    console.log("adding new containers per ACL listing");
     let currentdetails = [];
     if (currentsharing.includes(container)) {
       currentdetails = await client.getShareDetails(
--- a/swift_browser_ui_frontend/src/common/conv.js
+++ b/swift_browser_ui_frontend/src/common/conv.js
@@ -1,16 +1,17 @@
 
 import {
   getBucketMeta,
+  getAccessControlMeta,
 } from "./api";
 
-export default function getLangCookie () {
+export default function getLangCookie() {
   let matches = document.cookie.match(new RegExp(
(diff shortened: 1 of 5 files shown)


swift_browser_ui_frontend/src/common/conv.js
```diff
@@ -21,7 +23,108 @@ function shiftSizeDivision (vallist) {
...
+
+  // Delete stale shared container access entries from the database
+  for (let container of currentsharing) {
+    if (!(Object.keys(aclmeta).includes(container))) {
+      let resp = await client.shareContainerDeleteAccess(
+        project,
+        container,
+      );
+      if (!resp) {
+        console.log("Container access deletion failed");
```
csc-felipe, line 69: Did you mean to leave this here?

swift_browser_ui_frontend/src/common/conv.js
```diff
@@ -21,7 +23,108 @@ function shiftSizeDivision (vallist) {
...
+      }
+    } else {
```
csc-felipe, line 71: The `else` block seems to be unnecessary.

swift_browser_ui_frontend/src/entries/main.js
```diff
@@ -135,8 +141,20 @@ new Vue({
...
+      5000,
+    );
   },
   methods: {
+    containerSyncWrapper: function () {
+      syncContainerACLs(
+        this.$store.state.client,
+        this.$store.state.active.id,
+      ).then(() => {
+        console.log("ACLs have been updated.");
```
csc-felipe, line 155: Did you leave this intentionally?

csc-felipe (commented):
I'm not familiar with the ACLs yet, and I did not run it, as I haven't been running the whole environment yet. Thus, only looked at the code and made the review as a comment.

I would have preferred if the formatting had been a separate PR, as there's a lot of it.

swift_browser_ui_frontend/src/entries/main.js
sampsapenna: Somewhat intentionally – I'd like to flag the process somewhere, but a toast seems like overdoing it. It's true that console messages in production software are a bit unprofessional, though, so there's probably a better option somewhere. Maybe some sort of a loading indicator for background information could be useful for the UI.

I'll ditch the rest of the console messages.

swift_browser_ui_frontend/src/common/conv.js
sampsapenna: No, did not – good catch.

blankdots (approved):
if my understanding is correct this #69 is not required any more.

We need to add a follow-up issue to add some integration (or cypress) tests for this. A test could look like:
1. we log in we see some shared containers
2. we update the db in the background
3. we re-login the users we see the updated shared containers 

I have not tested, however the code looks good and if we spot any issues I assume we will see them in practice.

sampsapenna:
> if my understanding is correct this #69 is not required any more.
> 
> We need to add a follow-up issue to add some integration (or cypress) tests for this. A test could look like:
> 
> 1. we log in we see some shared containers
> 2. we update the db in the background
> 3. we re-login the users we see the updated shared containers
> 
> I have not tested, however the code looks good and if we spot any issues I assume we will see them in practice.
> 
> also this needs a rebase

Opened an issue for the integration tests in #424 

#69 is not relevant to this PR, but it would go nicely together with this change. This PR was about keeping the database information as current as possible. I'd like to keep it open still.

Rebase will happen shortly 👍🏼

sampsapenna:
> I'm not familiar with the ACLs yet, and I did not run it, as I haven't been running the whole environment yet. Thus, only looked at the code and made the review as a comment.
> 
> I would have preferred if the formatting had been a separate PR, as there's a lot of it.

Yup, the formatting was an unfortunate mishap with vscode auto-formatting that I noticed too late, thus it made its way in

swift_browser_ui_frontend/src/entries/main.js
sampsapenna: Removed the console messages, popped open an issue on adding a better loading indicator for the UI since we have begun to fetch more dynamic content in the background. Issue together with the discussion on it can be found in #425

PR #423, opened 2021-12-03
dev: persist sessions, shell command + vue devServer proxy.

csc-felipe:
### Description

This is hopefully helpful to others. What do you think?

**Persisting sessions** help so that when the python server stops and starts again, the active sessions are reloaded, no need to sign in every time.

**Shell command** adds the 'app' to the context of a python or iPython shell. Automating setting up the env to run certain commands.

**Vue devServer** uses the Webpack devServer to proxy requests to the locally running backend, so we can use 'npm run serve' to see our changes immediately, without having to run a production build after every change to the frontend.

### Type of change

### Testing

- [x] Tests do not apply

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,14 @@
+ipython
+readline
+tox
+pytest
+pytest-cov
+coverage
+flake8
+flake8-docstrings
+pytest-xdist
+asynctest
+black
+sphinx
+sphinx_rtd_theme
+pytest-timeout
--- a/swift_browser_ui/ui/server.py
+++ b/swift_browser_ui/ui/server.py
@@ -2,6 +2,8 @@
 
 # Generic imports
 import logging
+import os.path
+import pickle  # nosec
 import time
 import sys
 import asyncio
@@ -12,6 +14,7 @@ import secrets
 import uvloop
 import cryptography.fernet
 import aiohttp.web
+from aiohttp.web_exceptions import HTTPUnauthorized
 
 from swift_browser_ui.ui.front import index, browse, loginpassword
 from swift_browser_ui.ui.login import (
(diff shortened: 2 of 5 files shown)


requirements-dev.txt
```diff
@@ -0,0 +1,14 @@
+ipython
```
blankdots, line 1: why do we need ipython?

swift_browser_ui/ui/server.py
```diff
@@ -65,6 +72,22 @@
 async def kill_sess_on_shutdown(app: aiohttp.web.Application) -> None:
     """Kill all open sessions and purge their data when killed."""
     logging.info(f"Gracefully shutting down the program at {time.ctime()}")
+
+    if setd.get("set_session_devmode", False):
```
blankdots, line 76: i understand the utility, however this leaves room for error in config given that this is relevant for login, i think we need another way to do this

swift_browser_ui/ui/server.py
```diff
@@ -2,6 +2,8 @@
 
 # Generic imports
 import logging
+import os.path
+import pickle  # nosec
```
blankdots, line 6: importing packages with nosec not sure given that https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html it is imported even though not used

blankdots (commented):
not sure what to say, i understand the utility but this leaves too much room for errors in config and might introduce security concerns

swift_browser_ui/ui/server.py
csc-felipe: Might be better to put the import inside the block where it is used. The security issue is with the methods, not with the import.

swift_browser_ui/ui/server.py
csc-felipe: Yes, a dev server with auto reload would be better, and this would run only when running on the dev server, for example. I tried something on those lines, but didn't succeed yet. I'll try again some other time.

requirements-dev.txt
csc-felipe: For the `swift-browser-ui shell` command. It `await`s automatically. Nice to have, but not necessarily required. If it's not listed there, then need to remember to install it separately.

swift_browser_ui/ui/server.py
blankdots: fair enough.

swift_browser_ui/ui/server.py
blankdots: yes, probably we can get inspired from https://github.com/CSCfi/metadata-submitter/pull/287

csc-felipe:
Some of these changes already went in another PR. I'll make a new PR when there are other improvements, or when I manage to improve these enough.

PR #430, opened 2021-12-13
dev: devServer proxy, VSCode hints. 

csc-felipe:
### Description

- Uses the Webpack `devServer` to proxy requests to the locally running backend, so we can use `npm run serve` to see our changes immediately, without having to run a production build after every change to the frontend.
- Proper VSCode + Vetur hints with `jsconfig.json` and `vetur.config.js`.

### Type of change

### Changes Made
- Proxy the development frontend server to a locally running backend.
- `jsconfig.json` makes VSCode find code references and imports.
- `vetur.config.js` makes Vetur find the frontend code, as we use a monorepo.
- We actually don't have `npm run test`, so I removed it from the README.

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/README.md
+++ b/swift_browser_ui_frontend/README.md
@@ -6,6 +6,8 @@ npm install
 ```
 
 ### Compiles and hot-reloads for development
+This command also proxies the backend requests to a locally running server, by default `http://localhost:8080`.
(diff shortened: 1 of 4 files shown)


blankdots (approved):
i can see this being useful.

Issue #425, opened 2021-12-08
Indicator for background loading


sampsapenna:
#### Proposed solution

As a user I'd like to know if a background process is going on in the UI. For example, the tag fetches after the initial container fetch, sharing information update etc.

#### DoD (Definition of Done)

There's an indicator if something is loading in the background.

#### Testing

 A UI test can be added together with a long duration background load operation. The presence of this element during the load operation can be asserted in the frontend tests.

PR #435, opened 2021-12-22
Abort large number of fetch requests

csc-felipe:
### Description

Loading tags is unnecessary when navigating to another page, and it
keeps browser resources busy. This change cancels those requests when
browsing away.

### Type of change

- [x] New feature (non-breaking change which adds functionality

### Changes Made

Container and object view can cancel loading of tags when moving away of the page using an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController). 

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/common/api.js
+++ b/swift_browser_ui_frontend/src/common/api.js
@@ -77,6 +77,7 @@ export async function getBuckets() {
 
 export async function getBucketMeta(
   container,
+  signal,
 ){
   let url = new URL(
     "/api/bucket/meta?container=".concat(encodeURI(container)),
@@ -84,7 +85,7 @@ export async function getBucketMeta(
   );
 
   let ret = await fetch(
-    url, {method: "GET", credentials: "same-origin"},
+    url, {method: "GET", credentials: "same-origin", signal},
   );
   return ret.json();
 }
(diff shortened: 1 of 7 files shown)


blankdots (approved):
this sparks the idea of displaying a message, e.g. after 30 sec or so, that the loading time will take longer for this request and data is being processed and I assume this can be handled in a different PR.

Should we add this abortcontroller to `errorCaptured` https://vuejs.org/v2/api/#errorCaptured as well ?

otherwise seems ok.

csc-felipe:
There is an issue for showing long background ops #425. 

I'm not sure if I understood what do you mean with the `errorCaptured`.

sampsapenna (approved):
Solves the issue with existing fetch calls preventing updating the new route view when etc. opening a container, approved. The experience was considerably faster now, with projects that have hundreds of buckets.

blankdots:
> I'm not sure if I understood what do you mean with the `errorCaptured`.

Don't we need to abort the calls when an error is captured, or we will never have that case ?

csc-felipe:
> Don't we need to abort the calls when an error is captured, or we will never have that case ?

That would be good if we know what kind of errors might happen. I think first we should be able to observe what errors people are having, so we don't have to guess.

PR #436, opened 2021-12-22
Inform that container and objects can't be renamed.

csc-felipe:
### Description

Shows a meassage saying that container and objects can't be renamed. The tag featured introduced an edit view for container and objects, which may lead to the impression that somehow container and object names could be edited. This message tries to clarify that they can't.

@sampsapenna, please check the Finnish words. :)

image

### Type of change

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/common/lang.js
+++ b/swift_browser_ui_frontend/src/common/lang.js
@@ -197,6 +197,8 @@ let default_translations = {
       container_ops: {
         addContainer: "Add a new bucket",
         editContainer: "Editing bucket: ",
+        norename: "It is not possible to rename buckets. " +
+        "However, you can copy them with a new name.",
         deleteConfirm: "Delete Bucket",
         deleteConfirmMessage: "Are you sure you want to delete this " +
                               "bucket?",
@@ -209,6 +211,8 @@ let default_translations = {
       objects: {
         objectName: "Object",
         editObject: "Editing object: ",
+        norename: "It is not possible to rename objects. " +
+        "However, you can copy them with a new name.",
         deleteConfirm: "Delete Objects",
         deleteObjects: "Delete Object / Objects",
         deleteSuccess: "Objects deleted",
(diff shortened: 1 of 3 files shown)


swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -197,6 +197,7 @@ let default_translations = {
       container_ops: {
         addContainer: "Add a new bucket",
         editContainer: "Editing bucket: ",
+        norename: "It is not possible to rename buckets.",
```
blankdots, line 200:
```suggestion
        norename: "At the moment renaming buckets is not possible.",
```

blankdots (approved):
would phrase it differently, but other than that LGTM

swift_browser_ui_frontend/src/common/lang.js
csc-felipe: The rephrasing gives the impression that it is a planned feature. I had the impression that it is not. Did I understand wrongly?

blankdots: It is not a planned feature, however I think we will get a request for it, at some point.
We can phrase it differently than what i suggested, however what i wanted to express was that it is not supported, but it is possible as users are able to do copy & delete operations to get a new bucket/object.

blankdots: `We do not support renaming buckets/objects.` could be an alternative.

csc-felipe: A short explanation could be `It is not possible to rename container/object. However, you can copy it with a new name.`

blankdots: sure, that works

csc-felipe:
@sampsapenna, could you check the language, please?

This is how it looks like now.
image
image

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -439,6 +445,8 @@ let default_translations = {
       objects: {
         objectName: "Objekti",
         editObject: "Muokataan objekti: ",
+        norename: "Objekteja ei voi nimetä uudelleen. " +
```
sampsapenna, line 448: Probably best to move from `Objekteja` to `Objektia`. The following line is not in plural, and having it in singular would match the Finnish translation for Bucket edits.

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -439,6 +445,8 @@ let default_translations = {
...
+        "Voit kuitenkin kopioida sen uudella nimellä.",
```
sampsapenna, line 449: Same here, suggesting `Objektia ei voi nimetä uudelleen, mutta sen voi kopioida uudella nimellä.`

sampsapenna (dismissed):
Some minor alterations to the Finnish language, otherwise LGTM.

sampsapenna (approved):

PR #437, opened 2021-12-30
don't change state after application started

blankdots:
### Description

corresponds to: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_app.py#L146 and https://docs.aiohttp.org/en/stable/web_advanced.html#background-tasks

### Type of change

### Changes Made

1. changing state after application is not possible any more adding the states to `on_startup` for the `web.Application`
2. adjust unit tests
3. remove tox envs that are not used for py37 and py38

### Testing

- [x] Tests do not apply

### Mentions

Pending Testing ... to see if it behaves as it should

### DIFF
--- a/swift_browser_ui/ui/server.py
+++ b/swift_browser_ui/ui/server.py
@@ -64,6 +64,22 @@ from swift_browser_ui.ui._convenience import clear_session_info
 asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
 
 
+async def startup(app: aiohttp.web.Application) -> None:
+    """Add startup web server state configuration."""
+    # Mutable_map handles cookie storage, also stores the object that provides
+    # the encryption we use
+    app["Crypt"] = cryptography.fernet.Fernet(cryptography.fernet.Fernet.generate_key())
+    # Create a signature salt to prevent editing the signature on the client
+    # side. Hash function doesn't need to be cryptographically secure, it's
+    # just a convenient way of getting ascii output from byte values.
+    app["Salt"] = secrets.token_hex(64)
+    # Set application specific logging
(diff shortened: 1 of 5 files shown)


swift_browser_ui/ui/server.py
```diff
@@ -89,18 +105,7 @@ async def servinit() -> aiohttp.web.Application:
...
-    # Create a signature salt to prevent editing the signature on the client
-    # side. Hash function doesn't need to be cryptographically secure, it's
-    # just a convenient way of getting ascii output from byte values.
-    app["Salt"] = secrets.token_hex(64)
-    # Set application specific logging
-    app["Log"] = logging.getLogger("swift-browser-ui")
-    app["Log"].info("Set up logging for the swift-browser-ui application")
-    # Cookie keyed dict for storing session data
-    app["Sessions"] = {}
+    app.on_startup.append(startup)
```
csc-felipe, line 108: Maybe the startup could go together with the others at the end of the function.

tests/request/test_server.py
```diff
@@ -10,45 +9,23 @@
...
-    async def test_resume_on_start(self):
-        """Test resume on start function."""
-        await resume_on_start(self.mock_application)
-        self.mock_db_conn.open.assert_awaited_once()
 
-    async def test_graceful_shutdown(self):
-        """Test graceful shutdown function."""
-        await graceful_shutdown(self.mock_application)
-        self.mock_db_conn.close.assert_awaited_once()
+        pass
```
csc-felipe, line 23: It looks like the whole `setUp` function could be removed, no?

csc-felipe (approved):

Issue #22, opened 2019-09-17
Unit (+other) tests for the frontend


sampsapenna:
#### Description
As the frontend is now created using Vue-CLI, it can be unit tested with ease.
The unit (and other if required) tests for the frontend should be written.

#### DoD (Definition of Done)
The frontend is adequately tested.

#### Testing
Running the newly written unit (+other) tests after they have been written.

blankdots:
also introduce typescript

csc-felipe:
being tracked somewhere else

PR #438, opened 2021-12-30
remove asynctest & unittest_run_loop as deprecated

blankdots:
### Description

adjust for:
- `DeprecationWarning: Decorator "@unittest_run_loop" is no longer needed in aiohttp 3.8+`
- `DeprecationWarning: "@coroutine" decorator is deprecated since Python 3.8, use "async def" instead` from `asynctest` library

### Related issues

Part of #22 

### Type of change

- [x] refactor unit tests

### Changes Made

1. remove `asynctest` in favour of using unittest library Async functionality
2. other small adjustments

### Testing

- [x] Unit Tests


csc-felipe (approved):
This is nice, no more warnings and one less dependency.

PR #442, opened 2021-12-30
Fix operations with large objects in shared containers.

### LINKED ISSUE #388: Shared uploads can't handle a non-existent _segments container
#### Describe the bug

Current upload behavior is to break >5GiB files into segments as per Swift API guidelines. The process, however, currently expects that the segments should be pushed into a $CONTAINER_NAME_segments -container, which is a behavior similar to the native Openstack tools for object storage. The issue is, when a user has shared a container to the uploading party, the _segments container doesn't necessarily exist, thus leading to a situation where the upload runner creates a _segments container to the uploading project, instead of the _segments container being owned by the project that originally shared the container.

#### To Reproduce
1. Open the UI
2. Create a container and share that to another project
3. Using the UI with another project, upload a file larger than 5GiB to the container shared to the project in the previous step
4. Observe the _segments container being present in the uploading project, instead of the project that originally created the container

#### Expected behavior

All object contents should at least be accessible to the project that originally shared the container, either by sharing the segments container to the project that owns this shared container, or by uploading the segments into the shared container.

csc-felipe:
### Description

To fix the issues with operations on shared containers, we
upload large objects to the same container. Instead of using a
container for segments, they go to a pseudo-folder.

### Related issues
Fixes #388

### Type of change

### Changes Made

- Upload segments to the same container under `.segments/`
- Upload regular objects to `data/`
- Hide segments in the UI
- Calculate size of large objects
- Container downloads ignore the segments folder

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui/upload/common.py
+++ b/swift_browser_ui/upload/common.py
@@ -9,6 +9,10 @@ import keystoneauth1.session
 from swift_browser_ui.upload import upload
 
 
+DATA_PREFIX = "data/"
+SEGMENTS_PREFIX = ".segments/"
+
+
 def generate_download_url(
     host: str,
     container: typing.Union[str, None] = None,
--- a/swift_browser_ui/upload/download.py
+++ b/swift_browser_ui/upload/download.py
@@ -20,6 +20,7 @@ from swift_browser_ui.upload.common import (
     generate_download_url,
     get_path_from_list,
     get_download_host,
+    SEGMENTS_PREFIX,
 )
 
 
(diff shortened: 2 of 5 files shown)


csc-felipe:
Downloading the whole container will download segments as well, duplicating the large objects.

csc-felipe:
@sampsapenna, what other tools come to mind that should be updated regarding this change? Is there anything else we should take into consideration?

blankdots (approved):

PR #444, opened 2021-12-31
use pyspelling as spellchecker

blankdots:
### Description

this worked really well when introduced in https://github.com/CSCfi/metadata-submitter/pull/316 so we should adopt it.

### Type of change

- [x] Refactor spelling CI

### Changes Made

1. add dictionary of files
2. check docs and all markdown files
3. adjust some spelling
4. introduce `pyspelling` in github action as a improved alternative to `misspell` which is not that maintained
5. spellcheck lang.js file for English (1 mistake found) 

### Testing

- [x] Tests do not apply

### Mentions

I could not figure out the regex how to make it work https://facelessuser.github.io/pyspelling/filters/context/ to filter out just the `fi:` portion of the `lang.js` file (fi:((.|\n)*)` worked in other regex testers but not with the pyspelling - would appreciate some help if anyone has any ideas.


csc-felipe:
Did you already run it against the code to see what happens?

blankdots:
> Did you already run it against the code to see what happens?

no, not yet, will try next week to see what happens

csc-felipe:
Should we have the new files without the dot and inside the `.github` folder, if possible? It would make it more clear that they are only used for GH actions.

blankdots:
> Should we have the new files without the dot and inside the `.github` folder, if possible? It would make it more clear that they are only used for GH actions.

@csc-felipe done

csc-felipe (approved):
LGTM

blankdots:
merging this as it does not introduce any breaking changes, just spellchecking

PR #459, opened 2022-01-05
frontend: Fix errors for undefined variables

csc-felipe:
### Description

This fixes errors that show in console due to vuex initial state having `undefined` variables.

### Type of change

### Changes Made

- Set values for variables that would otherwise be `undefined`.
- Refactor `truncate()`, using only one implementation, and avoiding `undefined` errors.
- Initialized the translation library for the login page.

### Testing

- [x] Tests do not apply


blankdots:
if I understand what this does, that would be awesome to have, still looks like the integration tests fail for some reason, maybe those need adjusting or fixing :thinking:

blankdots (approved):

PR #476, opened 2022-01-27
Bugfix/cors headers

blankdots:
### Description

CORS errors in the console were annoying, trying to get rid of them.

### Type of change

- [x] Package updates
### Changes Made

- bump npm version to 8.3.2
- update node js to 14.8.3 and alpine 15
- remove console.logs
- add cors to errors in middleware
- move `404` to `204` when chunk not yet uploade as recommended by: https://github.com/23/resumable.js#handling-get-or-test-requests

### Testing

- [x] Tests do not apply

### Mentions

i personally would have liked to keep it `404` but that was very tricky to catch, if anyone has any better options feel free to create a PR.

~i am conflicted on why we use `resp.status not in {201, 202}` and not `resp.status not in [201, 202]` :thinking:~

### DIFF
--- a/.github/workflows/e2etests.yml
+++ b/.github/workflows/e2etests.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Build Frontend
         run: |
           pushd swift_browser_ui_frontend
-          sudo npm install -g npm@8.3.0
+          sudo npm install -g npm@8.3.2
           npm install
           npm run build
       - name: Setup and Run backend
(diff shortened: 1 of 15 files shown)


csc-felipe (approved):
Nice to have a fix for these errors.

About checking if an item is in a set vs if it's in a list, I've seen the pattern in other projects as well. I don't remember all the reasons, but according to this thread, there's a performance gain https://stackoverflow.com/questions/2831212/python-sets-vs-lists.

blankdots:
> About checking if an item is in a set vs if it's in a list, I've seen the pattern in other projects as well. I don't remember all the reasons, but according to this thread, there's a performance gain https://stackoverflow.com/questions/2831212/python-sets-vs-lists.
yes @sampsapenna was kind enough to explain that to me as well. now I got it.

csc-felipe:
I would be good to update the Changelog in this PR :)

PR #491, opened 2022-02-02
Error catching and hiding

### LINKED ISSUE #461: Better error handling in the frontend
#### Description
Network errors are not caught, and end up in the browser console, even if the error was expected. It happens, for e.g. for cancelled requests and API responses 4xx.

#### DoD (Definition of Done)
Network errors are properly handled and do not show up in the console. If relevant, the user should be informed of the error.

csc-felipe:
### Description
Adds code to catch and hide errors that usually show up in the console. Also adds the possibility to log to console on development builds.

### Related issues
Fixes #461.

### Type of change

### Changes Made

- Log errors to console only on dev builds
- Redirect to front page on session expired (HTTP 401)

### Testing

- [x] Tests do not apply

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - **BREAKING** Rename code occurrences of *bucket* to *container* (GH #471)
 - Updated dependencies in front-end as well as node.js base image to `node:14.18.3-alpine3.15`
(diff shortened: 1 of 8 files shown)


blankdots (commented):
I think we need to add the option for the devel containers to `build-devel` namely:
- https://github.com/CSCfi/swift-browser-ui/blob/master/dockerfiles/Dockerfile-build-crypt-devel
- https://github.com/CSCfi/swift-browser-ui/blob/master/dockerfiles/Dockerfile-ui-devel

blankdots (approved):

Issue #261, opened 2021-07-05
simultaneous browsing of all available projects


sampsapenna:
#### Proposed solution

User experience would improve if all the user's projects would be listed simultaneously. This means that the containers from different projects should be possible do display at the same time. This would also mean that the project in which each container is created should be conveyed to the user in some other manner.

#### DoD (Definition of Done)

Project no longer needs to be selected in order to display the container listing, but all containers in all projects are listed at once.

#### Testing

Unit and integration tests can be altered and extended to test for this new behavior.

sampsapenna:
Simultaneously browsing all projects would require scoping a separate token for each project, instead of having only one scoped token at any time. This would have the benefit of removing the need for token re-scoping upon project change, which would in turn improve the user experience.

PR #496, opened 2022-02-08
Replace synchronous Openstack Swift and Keystone libraries with asynchronous implementation

### LINKED ISSUE #259: implement paging in API requests
#### Proposed solution

Large requests in the backend API should execute quicker, since it can often take > minute for a large request to go through. This is due to large requests being pooled together in the backend, instead of the frontend – a behavior which could be improved upon by proxying paged requests and parsing them into a complete one in the frontend instead.

#### DoD (Definition of Done)

Paged requests are used in optimizing the container and object listings in the application.

#### Testing

New unit tests need to be written to reflect this functionality, which differs from the current implementation somewhat.

### LINKED ISSUE #260: move away from synchronous openstack library
#### Proposed solution

The library provided by Openstack is highly threaded synchronous library, which fits poorly in this use case. An asynchronous solution should be found instead to improve performance, especially during large object listings and authentication which can take up a long time.

#### DoD (Definition of Done)

Openstack API queries are performed asynchronously.

#### Testing

Unit tests should be refactored to reflect the asynchronous implementation of Openstack.

### LINKED ISSUE #440: Backend has access to all projects
#### Proposed solution
The backend should have access to all projects a user has access to, simultaneously. This would allow #261, to be implemented.

#### DoD (Definition of Done)
The backend can query, edit and create containers and objects of any project the user has access to, without re-scoping.

#### Testing
Peer review, system tests.

sampsapenna:
### Description

PR includes full async rewrite of `python-swiftclient` and `keystoneauth1` specifically tailored for our purposes, together with its implementation in the UI and upload backends. Container download still has some changes left to be made.

Generates a follow-up issue #501 to update API docs after major changes deprecation.

### Related issues

* Fixes #260  by deprecating synchronous Openstack libraries
* Fixes #259  by adding request paging to frontend when browsing large containers
* Prerequisite to #261  by removing forced project level scoping in the backend.
* Fixes #440 by changing the session to include scoped tokens for every project

### Type of change

- [x] This change requires a documentation update (this needs a follow up PR)

### Changes Made

* Remove backend dependency on synchronous Openstack native clients
* Initial prune of codebase
* Remove some duplicate backend functionality
* Add an asynchronous implementation of Openstack Swift and Openstack Keystone fit for our pupose of fast proxying
* Add own implementation of `bulk-delete` and `bulk-head` requests
* Move project scope to API routing, to prevent backend scope requirements
* Deprecate concatenating object and container listings in backend, now done in frontend
* Rewrite much of the test base for upload, UI backends
* Write new frontend test server as an aiohttp middleware

### Testing

- [x] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply


swift_browser_ui_frontend/src/views/Upload.vue
```diff
@@ -271,8 +273,8 @@ export default {
   },
   methods: {
     getPubKey: function () {
-      if (this.$t("message.keys")) {
-        for (let item of Object.entries(this.$t("message.keys"))) {
+      for (let item of Object.entries(this.$t("message.keys"))) {
```
blankdots, line 276:
```suggestion
      for (let item of Object.entries(this.$te("message.keys"))) {
```

swift_browser_ui/ui/login.py
```diff
@@ -157,169 +172,77 @@ async def sso_query_end(
...
-    }
-
-    # Save time of login to the backend
-    created = time.time()
-    request.app["Sessions"][session]["last_used"] = created
-    request.app["Sessions"][session]["max_lifetime"] = created + int(
-        setd["session_lifetime"]  # type: ignore
+    response = aiohttp.web.Response(
+        status=303,
+        body="",
```
blankdots, line 184: weird that it needs the empty body

swift_browser_ui/ui/login.py
```diff
@@ -332,18 +255,25 @@ async def login_with_token(
...
+                    headers={
+                        "X-Auth-Token": session["token"],
+                        "X-Subject-Token": session["projects"][project]["token"],
+                    },
+                ):
+                    pass
+            session.invalidate()
         except aiohttp.web.HTTPUnauthorized:
-            log.info(f"Trying to log our an invalidated session: {session}")
+            log.info(f"Trying to log our an invalidated session")
```
blankdots, line 275:
```suggestion
            log.info("Trying to log our an invalidated session")
```

blankdots (commented):
small things for now, till I am able to test it, also fixing flake8 would help

swift_browser_ui_frontend/src/views/Upload.vue
blankdots: my bad this should not be like that.

swift_browser_ui_frontend/src/common/api.js
```diff
@@ -110,39 +89,55 @@ export async function updateBucketMeta(
...
-          "/api/object/dload?bucket=" + container +
-          "&objkey=" + ret[i]["name"]
-        );
-      }
-      return ret;
-    },
+export async function getObjects(
+  project,
+  container,
+  marker = "",
```
csc-felipe, line 95: What is the marker?

swift_browser_ui_frontend/src/common/store.js
```diff
@@ -145,78 +142,90 @@ const store = new Vuex.Store({
...
-        commit("updateContainers", ret);
-        commit("loading", false);
-      }).catch(() => {
+        });
         commit("loading", false);
+        if (containers.length > 0) {
+          commit("updateContainers", state.containerCache.concat(containers));
+          marker = containers.pop().name;
+        }
+      } while (containers.length > 0);
```
csc-felipe, line 167: I didn't get this. It looks like it is making an extra request for each container. Why does it do that?

swift_browser_ui_frontend/src/common/store.js
```diff
@@ -145,78 +142,90 @@ const store = new Vuex.Store({
...
-        commit(
-          "updateContainerTags", 
-          {containerName: container.name, tags},
-        );
-      });
+    updateContainerTags: function (
+      { commit },
+      {project, containers, signal},
+    ) {
+      if (containers.length < 10000) {
```
csc-felipe, line 180: Is there any case of people with that many containers? We would need many other optimizations in that case.

swift_browser_ui_frontend/src/common/store.js
```diff
@@ -145,78 +142,90 @@ const store = new Vuex.Store({
...
+        console.log("container tag update called.");
+        containers.map(async container => {
+          console.log("Updating tags for container ", container.name);
```
csc-felipe, line 183: Couple of lingering console calls here

swift_browser_ui_frontend/src/common/store.js
```diff
@@ -145,78 +142,90 @@ const store = new Vuex.Store({
...
     },
     updateObjects: async function (
-      { commit, dispatch, state }, 
-      {route, signal},
+      { commit, dispatch, state },
+      { route, signal },
     ) {
-      let container = route.params.container;
       commit("loading", true);
-      if (route.name == "SharedObjects") {
```
csc-felipe, line 179: What happened to the shared objects?

swift_browser_ui_frontend/src/components/BrowserNavbar.vue
```diff
@@ -160,6 +163,19 @@ export default {
         "; path=/; expires="
         + expiryDate.toUTCString();
     },
+    changeActive (item) {
+      this.$store.commit("setActive", item);
+      let newParams = Object.fromEntries(Object.entries(
+        this.$route.params,
+      ));
+      if (newParams.project != undefined) {
```
csc-felipe, line 171: Is this for routes that don't have a project in the params?

swift_browser_ui_frontend/src/entries/main.js
```diff
@@ -103,32 +104,45 @@ new Vue({
...
         }
-      });
+      } else {
+        active = projects[0];
+      }
+      this.$store.commit("setUname", user);
+      this.$store.commit("setProjects", projects);
+      this.$store.commit("setActive", active);
+      if (document.location.pathname == "/browse") {
+        console.log("Pushing route to the browser.");
```
csc-felipe, line 134: console call

swift_browser_ui_frontend/src/entries/main.js
```diff
@@ -103,32 +104,45 @@ new Vue({
...
+          name: "ContainersView",
+          params: {
+            project: active.id,
+            user: user,
+          },
+        });
+      }
+    };
+    initialize().then(() => {
+      console.log("Application initialized.");
```
csc-felipe, line 145: ^

swift_browser_ui/ui/login.py
sampsapenna: Got fixed after the rebase 👍🏼

swift_browser_ui_frontend/src/common/api.js
sampsapenna: It's for the Openstack API paging – the actual API can only provide 10k items at once per call, the marker is the latest item so the API knows where to continue. The library provided by Openstack does this by spinning up a request for every page in its own thread, leading to these kinds of long calls taking a lot of time to complete – so now we're doing it in the front-end.

swift_browser_ui_frontend/src/common/store.js
sampsapenna: "Null-terminated" listings. If there's nothing left after the marker, Swift API returns an empty listing signifying the end. It can't reliably be done by checking if the page contains less items than the page size, the page size is an upper limit and is not always an exact thing. Openstack's own implementation does it in a similar manner.

swift_browser_ui_frontend/src/common/store.js
sampsapenna: As far as I know it's a use case we should and need to support. Not unfeasible to expect >10k items in a container. But for now I think not being able to provide all the nice things we have added on top of the Openstack APIs when people reach the limits of the storage backend is a reasonable compromise.

swift_browser_ui_frontend/src/common/store.js
sampsapenna: Deprecated, since the object listing  is project aware now we no longer need a different endpoint for fetching over project borders.

swift_browser_ui_frontend/src/components/BrowserNavbar.vue
sampsapenna: Mainly for the project selection in the navbar, for additional reason because of the upload runner – it cannot depend on the route as the user should be able do navigate the UI while uploading things.

sampsapenna: Project selector could use the route project id just as well, though

csc-felipe:
I guess this fixes #440.

sampsapenna:
> I guess this fixes #440.

It does, thanks for noticing, I'll add it to the PR 👍🏼

blankdots (approved):
the cy.wait(1000) need to be done differently because cypress does not recommends waits to be used except some special circumstances, but these can be tackled in different issue

swift_browser_ui_frontend/src/components/ObjectDeleteButton.vue
```diff
@@ -56,22 +56,17 @@ export default {
         this.$store.state.db.objects.bulkDelete(objIDs);
       }
       swiftDeleteObjects(
+        this.$route.params.project,
         this.$route.params.container,
         to_remove,
-      ).then(() => {
-        if(this.$route.name === "SharedObjects") {
-          this.$store.dispatch(
-            "updateSharedObjects",
```
csc-felipe, line 64: I suspect that this one should have stayed. We only need to fetch objects again in the SharedObjects view, as it doesn't use the IDB, and now it is not refreshing the shared objects.

csc-felipe (approved):

PR #506, opened 2022-02-11
Search improvements

csc-felipe:
### Description
Performance and UX improvements for the search.

### Type of change

### Changes Made

- Search box in ObjectTable now reads 'Filter by...'
- Ranking of search results
- Highlight matching results
- exclude _segments containers from the results
- Better UX while typing
- Only query word searches idb, following ones filter in-memory
- Fix AddContainer not creating tokens for container
- Limit number of search results
- Remove apostrophe from search tokenizer

### Testing

- [x] Tests do not apply

### DIFF
--- a/.github/config/.finnishwords.txt
+++ b/.github/config/.finnishwords.txt
@@ -287,6 +287,7 @@ sivun
 sivupyynnössä
 sivutus
 sopimus
+suodata
 suuren
 suuret
 suuri
--- a/swift_browser_ui_frontend/src/components/SearchResultItem.vue
+++ b/swift_browser_ui_frontend/src/components/SearchResultItem.vue
@@ -4,11 +4,14 @@
   >
     <div class="media">
       <div class="media-content">
-        <span><b>{{ 
-          isContainer() 
-            ? $t('message.search.container')
-            : $t('message.search.object')
-        }}: </b> {{ filename() }}
+        <span>
+          <b>{{ 
+            isContainer() 
+              ? $t('message.search.container')
+              : $t('message.search.object')
+          }}: </b>
+          <!-- eslint-disable-next-line -->
+          <span v-html="getFilename()"></span>
         </span>
         <br>
         <small>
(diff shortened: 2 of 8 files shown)


swift_browser_ui_frontend/src/components/SearchResultItem.vue
```diff
@@ -18,12 +22,14 @@
           </span>
           <span v-if="hasPath()">
             <b>{{ $t('message.search.folder') }}: </b>
-            {{ filePath() }}
+            <!-- eslint-disable-next-line -->
+            <span v-html="getFilePath()"></span>
```
blankdots, line 26: do we really need to use `v-html` (not only here but all places) ?
We are using it on user content. I understand the content is to some degree safe, still, I am a bit uneasy using it. As an alternative we can use filters.

I assume eslint was disabled because of the XSS issue with `v-html` ?

.github/config/.finnishwords.txt
```diff
@@ -287,6 +287,7 @@ sivun
 sivupyynnössä
 sivutus
 sopimus
+Suodata
```
blankdots, line 290: lowercase should be fine.

swift_browser_ui_frontend/src/components/SearchResultItem.vue
csc-felipe: Yes. The user input is escaped before being processed and rendered, so I hope it would be ok :)

I tried using filters, but it doesn't render html. The only way to highlight, AFAIK, is with html. Another approach would be to convert the text to a structure format (array/object) and renders it as expected. Do you think this would be necessary, or the escaping is enough?

blankdots: > or the escaping is enough

enough i don't know, but I think we should take it one step forward to use https://www.npmjs.com/package/sanitize-html and limit the tags that can be used there. - I think we only need `<span>` from what i see, although `<b>`/`<strong>` could have worked as well.

csc-felipe: The user input has all non letter input removed, as part of the tokenization: https://github.com/CSCfi/swift-browser-ui/blob/devel/swift_browser_ui_frontend/src/common/conv.js#L245
AFAIK, it makes the input safe from client side XSS attacks.

csc-felipe: The tokenization is one of the first things that are done to the search query https://github.com/CSCfi/swift-browser-ui/pull/506/files#diff-4da6dd94ffeace796df8cb4bb9a6f9e0bdd9a8cadaa5779d2fb2b160d96fb02cR385

blankdots (approved):

PR #513, opened 2022-02-18
use redis cookie storage

sampsapenna:
### Description

We need too much session data to be stored into a session cookie, thus moving to using `aiohttp_session` with Redis support.

This change breaks compatibility when running in environments without redis configured. Session storage in Redis allows scaling the UI API properly.

### Type of change

### Changes Made

* Move from `EncryptedCookieStorage` to `RedisStorage`
* Configure necessary setup for opening a Redis connection

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui/ui/server.py
+++ b/swift_browser_ui/ui/server.py
@@ -2,6 +2,7 @@
 
 # Generic imports
 import logging
+import os
 import sys
 import asyncio
 import ssl
@@ -15,7 +16,9 @@ import uvloop
 import aiohttp.web
 
 import aiohttp_session
-import aiohttp_session.cookie_storage
+import aiohttp_session.redis_storage
+
+import aioredis
 
 from swift_browser_ui.ui.front import index, browse, loginpassword
 from swift_browser_ui.ui.login import (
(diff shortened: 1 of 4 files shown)


swift_browser_ui/ui/server.py
```diff
@@ -82,12 +85,24 @@ async def servinit(
     app = aiohttp.web.Application(middlewares=middlewares)  # type: ignore
 
     # Initialize aiohttp_session
+    redis_creds = ""
+    redis_user = str(os.environ.get("SWIFT_UI_REDIS_USER", ""))
```
blankdots, line 89: if we require these, there might be an option to omit the default so that server does not start if the variables are not set.

blankdots (approved):

swift_browser_ui/ui/server.py
sampsapenna: None of these are strictly required, a redis container with defaults will work when you just start the browser backend. These are to allow running on e.g. Rahti where credentials and host information are a must.

csc-felipe (approved):

PR #514, opened 2022-02-18
Run integration tests against CSCfi/docker-keystone-swift container.

### LINKED ISSUE #502: On network error, items are removed from IndexedDB
#### Describe the bug
When connection is lost, containers and objects may be removed from local cache.

#### To Reproduce
1. Go to a projects, wait that cache is loaded
2. Turn off wifi
3. Click something

#### Expected behavior
Items should not be removed on network error.

### LINKED ISSUE #510: Reduce flakiness in integration tests
#### Description
Cypress tests are flaking, let's improve that.

#### DoD (Definition of Done)
Tests are more stable :)

#### Testing
Tests pass or not as they should

csc-felipe:
### Description

Running integration tests against a container running [keystone and swift](https://github.com/CSCfi/docker-keystone-swift) instead of the mock server will allow us to have more comprehensive and realistic tests.

A few changes were made so that the tests work as expected.

Doesn't include wasm build, so encrypted uploads wouldn't work with the current setup.

### Related issues
- Fixes #502 
- Fixes #510

### Type of change

### Changes Made
- Github actions run integration tests with keystone and swift in a container.
- Fixed items being removed from IndexedDB on network errors.
- Fixed container and metadata updates.
- Make user roles configurable.

### Testing

- [x] Integration Tests


.github/workflows/e2etests.yml
```diff
@@ -8,25 +8,53 @@ jobs:
...
+      postgres:
+        image: postgres:alpine
+        env:
+          POSTGRES_PASSWORD: pass
+        ports:
+          - 5432:5432
+        volumes:
+          - ${{ github.workspace }}/init-project-db.sh:/docker-entrypoint-initdb.d/init-user-db.sh
+      keystone-swift:
+        image: ghcr.io/cscfi/keystone-swift:latest
```
blankdots, line 25: this repo, where is it?
there is no image here: https://github.com/CSCfi/docker-keystone-swift/packages, maybe you need to add https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages to that repository then you will be able to pull it with `ghcr.io/CSCfi/docker-keystone-swift`

sampsapenna (approved):

blankdots (approved):
instead of using honco and Procfile maybe having a docker-compose file with our services would have been also ok. like using: https://github.com/CSCfi/swift-ui-deployment/tree/master/integration

csc-felipe:
> instead of using honco and Procfile maybe having a docker-compose file with our services would have been also ok. like using: https://github.com/CSCfi/swift-ui-deployment/tree/master/integration

When we test encrypted upload we should do that, because the current setup doesn't have wasm build.

I wanted to keep it simple, so we don't have to build a docker image.

PR #520, opened 2022-02-22
Fix getting object metadata for large containers.

### LINKED ISSUE #504: Fetching object tags makes too long URL
#### Describe the bug
The object tag fetching logic is creating too long URLs.

`aiohttp.http_exceptions.LineTooLong: 400, message='Got more than 8190 bytes (8198) when reading Status line is too long.'`

#### To Reproduce
1. Go to browser view, in a project with containers that have a lot of objects
2. Look at the backend logs, look at console logs

#### Expected behavior
URLs should be smaller than 8190 bytes.

csc-felipe:
### Description
Prevents url from being longer than aiohttp's limit.

### Related issues
Fixes #504.

### Type of change

### Changes Made

- Checks if url will exceed maximum size, and make a request before that happens.

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/common/store.js
+++ b/swift_browser_ui_frontend/src/common/store.js
@@ -311,7 +311,7 @@ const store = new Vuex.Store({
 
       for (let i = 0; i < objects.length; i++) {
         // Object names end up in the URL, which has hard length limits.
-        // The aiohttp backend has a limit of 8192. The maximum size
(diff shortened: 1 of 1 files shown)


blankdots (approved):
a test for this would be nice.
also we can think to increase that byte size, if we ever find this to become a common occurrence.

approving as it seems to solve base issue.

sampsapenna (approved):

PR #521, opened 2022-02-23
invalidate expired sessions

sampsapenna:
### Description

Sessions that have an expired token should be properly checked for, not only for frontend requests. This is now handled in a middleware which is more aggressive about invalidating sessions that have expired.

### Type of change

### Changes Made

* Check for token expiration in middleware
* Remove token expiration checks elsewhere

### Testing

- [x] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply

### DIFF
--- a/swift_browser_ui/ui/middlewares.py
+++ b/swift_browser_ui/ui/middlewares.py
@@ -1,9 +1,11 @@
 """Middlewares for the swift-browser-ui."""
 
 
+import time
 import typing
 
 from aiohttp import web
+import aiohttp_session
 
 from swift_browser_ui.ui.settings import setd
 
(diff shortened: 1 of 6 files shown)


swift_browser_ui/ui/middlewares.py
```diff
@@ -27,6 +29,21 @@ def return_error_response(error_code: int) -> web.Response:
...
 
 
+@web.middleware
+async def check_session_at(
+    request: web.Request,
+    handler: AiohttpHandler,
+) -> web.Response:
+    """Raise on expired sessions."""
+    session = await aiohttp_session.get_session(request)
+    if "at" in session:
```
blankdots, line 39: we could also check the token and projects are present, or having a valid session is enough ?

sampsapenna: Valid session is enough, since the login process can't finish without the user having at least one project available, and a successful login to that project. That actually reminds me I need to remove some additional checks from the frontend.

blankdots:
Might want to consider adding, if we are doing stuff right `headers={"WWW-Authenticate": 'OAuth realm="/", charset="UTF-8"'}` to `HTTPUnauthorized` - required as specified by RFC https://httpstatuses.com/401

swift_browser_ui/ui/middlewares.py
```diff
@@ -27,6 +29,21 @@ def return_error_response(error_code: int) -> web.Response:
...
+        if session["at"] + 28800 < time.time():
```
blankdots, line 40:
```suggestion
    if "at" in session and session["at"] + 28800 < time.time():
```
I don't see any reason not do it like this also, or am I missing something ?

sampsapenna: Nope, that's fine – the comparison aborts if in fails so it won't generate a KeyError. I'll commit it 👍🏼

blankdots: maybe the indentation should have been fixed (my bad), but we can amend the last commit

PR #552, opened 2022-03-23
Optional two-step login with OIDC as first step.

### LINKED ISSUE #546: As a researcher, I want to be able to log in to SD Connect so that I can use the service I have applied access in MyCSC
#### Describe the user story

In order to use SD Connect, it will be necessary to apply access in MyCSC. The information about the user right to use the service is used to determine if user can access the service or not.  

We could re-implement the login phase so that user would need to log in in two different phases. This would enable us to consume service profile information from (SDS) AAI.  

#### Tasks:
- [x] https://github.com/CSCfi/swift-browser-ui/pull/552
- [x] https://github.com/CSCfi/swift-browser-ui/issues/571
- [x] https://github.com/CSCfi/swift-browser-ui/issues/601
- [x] https://github.com/CSCfi/swift-browser-ui/issues/565
- [x] #605
- [x] #606

#### Testing 
This will be manually tested in test environment before approval. Unit tests implemented if relevant. 

#### Testing Definition of Done
Given that I log-in with valid user account and applied right to use the service when I do log-in to SD Connect then I expect to get into the service

csc-felipe:
### Description
This change allows retrieving more user information
from the identity provider.

The frontend introduces a new index page for when
OpenID Connect (OIDC) is enabled. The new page uses
CSCfi/csc-ui components.

To use OIDC, new configuration variables must be set.
See `settings.py` and `vue.config.js`. 

To build the frontend, pass `OIDC_ENABLED=TRUE` to npm.

### Related issues
- #546.

### Type of change

### Changes Made

- New index page with CSC design when OIDC is enabled
- After OIDC login, user is redirected to the previously existing index page for the second login.
- Updated docker images to accept `OIDC_ENABLED` parameter, which should be set to `True` to build frontend with OIDC support.

### Testing

- [x] Unit Tests

### Mentions
Thanks to @teemukataja for the certificate fix and existing OIDC implementation, @villeerikssoncsc for the csc-ui support, @sampsapenna for helping to think the implementation.


dockerfiles/geant-ca
```diff
@@ -0,0 +1,39 @@
+-----BEGIN CERTIFICATE-----
```
blankdots, line 1: I would not place this here, this does not seem like a good place for us to have something like this in here. we can maybe use openssl commands to retrieve it or keep track of it, especially if it for development.

It can be mounted with a volume configuration for either test or production.

blankdots (changes requested):
the certificate has no place in the repository

blankdots:
This is not related to the code,
I am a bit conflicted of introducing this change, as based on the interaction it seems confusing for the users, and it does not seem to provide much in term of benefits of what data we can extract from the AAI. I am personally against moving along with this feature of 2 step authentication and waiting till we can do 1 step authentication.

juhtornr:
> This is not related to the code, I am a bit conflicted of introducing this change, as based on the interaction it seems confusing for the users, and it does not seem to provide much in term of benefits of what data we can extract from the AAI. I am personally against moving along with this feature of 2 step authentication and waiting till we can do 1 step authentication.
> 
> Or if we introduce this change, would like to see some user feedback (actually user testing) on how this performs with users and if it makes sense to introduce based on what we are looking to solve, by introducing it.

I see very few options to this change if we want to support service profile (#555) and enable proper access control to the service. Unless there are other options, we need to continue with this.

blankdots:
> > This is not related to the code, I am a bit conflicted of introducing this change, as based on the interaction it seems confusing for the users, and it does not seem to provide much in term of benefits of what data we can extract from the AAI. I am personally against moving along with this feature of 2 step authentication and waiting till we can do 1 step authentication.
> > 
> > Or if we introduce this change, would like to see some user feedback (actually user testing) on how this performs with users and if it makes sense to introduce based on what we are looking to solve, by introducing it.
> 
> I see very few options to this change if we want to support service profile (#555) and enable proper access control to the service. Unless there are other options, we need to continue with this. 

we do A/B testing and see if that yield what we expect

blankdots (changes requested):
I think there is a bug when adapting to CSC UI design components when i start the inspector tools in Firefox i get:

image

I suspect that comes from adapting to the csc design style guides, and i think this should be fixed.
Not sure what causes it the footer, maybe it needs to be in c-row, or that it is fixed or body missing 
```
 margin: 0px;
    max-width: 100vw;
```
or a combination of a lot of styles

dockerfiles/geant-ca
csc-felipe: I agree. I was wondering what would be the best way to solve this issue. We don't use that Dockerfile in production, but only in testing or local dev.

Then there should be a script as `Entrypoint`, which looks for a certificate, and if it's found, it's appended.

Or is it a good solution to pass the certificate as build `ARG`?

csc-felipe:
> Not sure what causes it the footer, maybe it needs to be in c-row, or that it is fixed or body missing

I see a couple of issues in your screenshot.
1. the background is only halfway. That is due to 100vh in `c-main`. @villeerikssoncsc, what do you think?
2. The footer got some content cut.

I have Firefox 91.8.0esr and Chromium 99, and I can't reproduce 2. Tried different ways, including lower screen resolution
image

I noticed 1 during development, but I didn't think it was such a big issue, though I agree it's better if the background covers the whole area.

blankdots:
> I noticed 1 during development, but I didn't think it was such a big issue, though I agree it's better if the background covers the whole area.

i think we should fix it with this PR, as we will be breaking a lot of stuff going further.

dockerfiles/geant-ca
blankdots: this does not seem like a build `ARG` to justify that, it seems like something we can mount with a volume or use with a normal environment variable.

blankdots:
> I have Firefox 91.8.0esr and Chromium 99, and I can't reproduce 2. Tried different ways, including lower screen resolution

to reproduce the page needs to be small enough to enable scroll, and I was able to reproduce in both Firefox and Chrome
Peek 2022-04-13 08-24

csc-felipe:
> i think we should fix it with this PR, as we will be breaking a lot of stuff going further.

What will break?

Ah, you mentioned the footer, but it's only the background you were talking about.

That is not related to the footer. `c-main`'s background has `height: 100vh`, so it is only as tall as the visible window. I guess it was made with the assumption that the screen is not scrollable.

I don't know how to fix this the proper way, as it's part of the component library. Let's see what Ville says. Possibly I'm using the library wrongly.

blankdots:
> What will break?

styles, have a tendency to cascade and if those are not fixed or if the html structure does not fit what we introduce, it might introduce issues we cannot be aware of at the moment, so imo it might be worth to take the time to do it properly, without leaving these kind of issues.

blankdots:
> Also this is not a library issue, i noticed it myself when i was adapting a service to these components, and i solved by re-arranging the html structure and css

for example with using the csc design components the fixed footer does not make sense, that can be done in a different manner

csc-felipe:
> for example with using the csc design components the fixed footer does not make sense used that way, that can be done in a different manner

I am open to suggestions, of course. How do you think is better to do it?
I made it so that it looks and works as similar to metadata-submitter and sd-desktop. If we do it differently, then the design will deviate from the others.


> Also this is not a library issue, i noticed it myself when i was adapting a service to these components, and i solved by re-arranging the html structure and css

I have tried many arrangements before, just tried a few more, and did not succeed. I can keep spending more time with this, but I would rather hear from someone who knows how to use the CSC library.

dockerfiles/geant-ca
csc-felipe: I don't see how to append the certificate to the existing file with your suggestion. Do you have an example?

blankdots:
> I am open to suggestions, of course. How do you think is better to do it?
> I made it so that it looks and works as similar to metadata-submitter and sd-desktop. If we do it differently, then the design will deviate from the others.

well maybe update to the latest library of csc design system, the version i used seems to be newer,

a simple fix would be to remove `c-main's background has height: 100vh` that seems to fix it at least from my tests, the rest of suggestions in the chat

dockerfiles/geant-ca
csc-felipe: Removed the cert

blankdots: example, the `requests.get` can be a path or a volume, and it can be done in a different stage of the docket build
```
customca = requests.get('http://place-to-download-your-root-ca-pem.file').content 
cafile = certifi.where() 
with open(cafile, 'ab') as outfile: 
    outfile.write(b'\n')
    outfile.write(customca)
```

blankdots: you can also use the same method you used with `cat` and `>>`, however you load the certificate from a build arg as you initially said, i am not sure i have a solution with env var :thinking:

csc-felipe: Your example above is what I had in mind for env var.

csc-felipe: The cert is not necessary, after all. There was a misconfiguration in the OIDC provider that the intermediary cert was not provided.

genie9:
@csc-felipe does this fix the whole user story https://github.com/CSCfi/swift-browser-ui/issues/546 or some of its tasks?

sampsapenna (approved):

csc-felipe:
> @csc-felipe does this fix the whole user story #546 or some of its tasks?

@genie9, only the first task, as can be seen from the issue. 
The issue was updated with more tasks after I created the PR, and the PR still had the `Fixes` in the PR description, so I removed it now to clarify.

PR #568, opened 2022-04-08
Not all paths should be proxied.

csc-felipe:
### Description

Some paths are handled in the frontend, and shouldn't be proxied,
so the proxied paths need to be set explicitly.

Also ws server was configured with backend server parameters, but
it should be configured with frontend parameters.

hot doesn't need to be set. It works like that by default.

The development server can be run with parameters, as
`PORT=8081 BACKEND_PORT=8001 npm --prefix swift_browser_ui_frontend/ run serve`
The default for `PORT` is `8080`, so it doesn't need to be set in that case.

### Type of change

### Changes Made

- Limit the URL paths that are proxied
- Fix ws server parameters

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/vue.config.js
+++ b/swift_browser_ui_frontend/vue.config.js
@@ -1,12 +1,25 @@
+const proxyTo = `http://${process.env.BACKEND_HOST || "localhost"}:${process.env.BACKEND_PORT || "8080"}`;
+
 module.exports = {  // eslint-disable-line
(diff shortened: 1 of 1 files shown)


blankdots (approved):

blankdots:
seems the integration tests don't like the change :(

csc-felipe:
I'll make another PR with more changes.

PR #600, opened 2022-05-04
docker with OIDC does need permission in app/

csc-felipe:
### Description

Building docker images with OIDC support actually needs access to `app/` folder, as 
it creates files. I had removed it thinking that it was because of the certificate, which was then not included.

Follow up from #552.

### Type of change

### Changes Made

- add chown to dockerfiles for building with OIDC  support

### Testing

- [x] Tests do not apply


sampsapenna (approved):
LGTM 👍🏼 Since it doesn't cause any conflicts with the base branch it can be merged immediately. Sorry for taking so long with this.

PR #607, opened 2022-05-17
Skip second login button click when first authentication used Haka

### LINKED ISSUE #565: Skip second login button click when using Haka as first login step
#### Proposed solution

At the moment, when user logs in with HAKA he/she is redirected to another login page where he/she needs to press login button again. This step can be skipped with utilising existing HAKA login session.

csc-felipe:
### Description

There is a new OIDC claim which indicates which authentication method was chosen. It comes as `homeFederation` from the provider.

With this, it's possible to skip the second step login button click, when the first one used Haka, as the second step also uses Haka.

### Related issues

Closes #565 

### Type of change

### Changes Made

- Skip second login button click when first authentication used Haka

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui/ui/login.py
+++ b/swift_browser_ui/ui/login.py
@@ -21,6 +21,12 @@ from swift_browser_ui.ui._convenience import (
 from swift_browser_ui.ui.settings import setd
 
 
+HAKA_ENDPOINT = (
+    "{endpoint}/auth/OS-FEDERATION/identity_providers"
+    "/haka/protocols/saml2/websso?origin={origin}"
+).format
+
+
 async def oidc_start(request: aiohttp.web.Request) -> aiohttp.web.Response:
     """Redirect to OpenID Connect provider."""
     try:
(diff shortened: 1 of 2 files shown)


swift_browser_ui/ui/login.py
```diff
@@ -80,11 +86,20 @@ async def oidc_end(request: aiohttp.web.Request) -> aiohttp.web.Response:
         "state": oidc_result["state"],
         "access_token": oidc_result["token"],
     }
+    request.app["Log"].info(session["oidc"])
```
csc-felipe, line 89: This should be removed, but it's the way to see the content from the OIDC provider.

sampsapenna: Ok, though it might be fit to lower the logging level to `debug`

sampsapenna (approved):

csc-felipe:
> [Ok, though it might be fit to lower the logging level to debug](https://github.com/CSCfi/swift-browser-ui/pull/607#discussion_r876683479)

Changed it to `debug`.

sampsapenna (approved):

PR #646, opened 2022-06-22
Update all deps

csc-felipe:
### Description

Update all deps. 
Cypress got a major version update, and apparently tests are failing also because of that, so also updating the configs to match new version.

### Type of change

### Changes Made

- Update frontend deps
- Update backend deps - Reordered deps to be in alphabetic order
- Update github actions
- Update node docker image
- Update cypress config for v10

### Testing

- [x] Tests do not apply


sampsapenna (dismissed):
The cypress updates are lacking a bump in cypress GitHub action, `v10` requires action `v4`. We already have that in place in other branches affecting UI, e.g. `feature/new-error-pages` which is getting merged soon. That's why e2e is still failing.

If it's not too much to ask, could you rebase on that one, since the tests have changed a little as well?

Other than that LGTM, and thanks!

sampsapenna (approved):

Issue #611, opened 2022-05-30
Upload


ainoc:
Given that user wants to upload files
When user selects upload button
Then user will see upload modal and can start upload process.

Given that user wants to upload
When user fills folder name field
and selects files 
then user will see of files underneath select files box 
and then user can select Upload button.

- [x] #648

- [x] #613

PR #674, opened 2022-07-14
Feature/upload modal

hannyle:
### Description

Implementing a new modal for uploading files based on the new UI design. 
https://www.figma.com/file/N0Y3N0nfyMS2wG3gjwZ201/SD-Connect-Prototype?node-id=604%3A6706

### Related issues

Closes https://github.com/CSCfi/swift-browser-ui/issues/611

### Type of change

### Changes Made

- Removed the routes to upload files
- Added global state for toggling Upload modal
- Restructured the components inside Upload modal
- Switched to use csc-modal and csc-data-table
- Moved the full-page drag and drop area inside the Upload modal
- Removed some functions which are not supposed to exist in the new UI, for example: **privateKey**, **publicKey**, **encryptAndUpload** (_please correct me if I understand it wrong_)
- Updated Changelog

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply, because I would like to make sure first that this feature works in a way as it should.


sampsapenna:
> My two cents about things I'm still not sure about this UI feature:
> 
> * When a user types in a folder name, how should the tag field show? If the existing folder already has some tags, how do we know when to render their tags, assuming the user can type back and forth some similar folder names.
> * I have removed some key functions which are not supposed to exist in the new UI, for example: **_privateKey_**, **_publicKey_**, **_encryptAndUpload_** but since I haven't tested with real example, please correct me if there is something supposed to work differently.

Existing tags can be checked from the `IndexedDB` whenever required – the database is quite fast, so polling it whenever bucket name changes shouldn't be an issue.

The key functions shouldn't be removed, since depending on whether we're deploying **SD Connect** or **Allas** we want encryption on or off by default. **SD Connect** uploads use the `encryptAndUpload` upload chain by default, **Allas** uses the normal upload. Private key functions can probably go, since we're not supporting those in production.

So, what the upload button runs depends on context. Both contexts need to be supported, but can be handled in the background. You also seem to have gotten rid of upload viability changes (logic that handles updating the `noUpload` flag). You can add those back if you want, though I think I can just add the updated versions while rebasing on your branch.

swift_browser_ui_frontend/src/views/Upload.vue
```diff
@@ -626,56 +308,112 @@ export default {
...
-  margin: auto;
+<style lang="scss" scoped>
+@import "@/css/prod.scss";
+
+.upload-card {
+  width: 64vw;
+  padding: 3rem;
+  left: 50%;
+  margin-top: 50%;
+  transform: translate(-50%, -50%);
```
sampsapenna, line 360: Translating this way seems to push the modal out-of-bounds on my machine, not sure if it's a low DPI monitor issue or what. Happens on both Firefox and Chrome, and with multiple different zoom levels.

SD_Connect_OOB_Chrome
SD_Connect_OOB_Firefox

swift_browser_ui_frontend/src/views/Upload.vue
```diff
@@ -1,289 +1,109 @@
 <template>
```
sampsapenna, line 1: `swift_browser_ui_frontend/src/views/Upload.vue` should be moved to `swift_browser_ui_frontend/src/components/Upload.vue`, and probably renamed to `UploadModal.vue` to adhere to vue naming convention. `views` folder is reserved for routing root views, and upload functions as a component now. This should've probably been documented, I'll make a PR out of it afterwards.

swift_browser_ui_frontend/src/views/Upload.vue
sampsapenna: I'll double check it's not due to having dev console open. I don't remeber it being so, but it seems to cause this for other modals

saulipurhonen:
One thing I noted is that modal width is quite narrow compared to prototype. 600px vs 920px.

hannyle:
> Is it possible to display scroll bar always if content doesn't fit in view?

--> I'm not sure what you mean. Does the scroll bar not show for you in certain views?

> I think scroll bar shouldn't be on top of content. Is it possible to move it next to the edge of modal?

--> I think it's possible. Next thing for me to fix.

saulipurhonen:
> > Is it possible to display scroll bar always if content doesn't fit in view?
> 
> --> I'm not sure what you mean. Does the scroll bar not show for you in certain views?

I initially tested with Chrome on Mac and it doesn't display scrollbar before I scroll. Firefox seems to work fine.

Here's a screenshot from Chrome:
image

And here's from FF:
image

saulipurhonen:
I'm not sure if this should be part of this PR but I tried enabling upload button manually and I think it would make sense to close the modal when encryption & upload begins. Now it stays open and a notification is rendered behind the modal background overlay like this:

image

swift_browser_ui_frontend/src/views/Containers.vue
```diff
@@ -303,8 +360,17 @@ export default {
     active() {
       return this.$store.state.active;
     },
-    openCreateFolderModal() {
-      return this.$store.state.openCreateFolderModal;
+    openCreateFolderModal: {
+      get() {
+        return this.$store.state.openCreateFolderModal;
+      },
+      set() {},
```
saulipurhonen, line 367: Is empty setter needed? I don't mind it being there but just wondering if it has a meaning.

hannyle:
> I'm not sure if this should be part of this PR but I tried enabling upload button manually and I think it would make sense to close the modal when encryption & upload begins. Now it stays open and a notification is rendered behind the modal background overlay like this:
> 
> image

Sure, closing the modal is quick to implement. I will do it.

swift_browser_ui_frontend/src/views/Containers.vue
hannyle: I had to add **get** and **set** there because without them, I had the warnings from console that those methods were lacking. I haven't tried other solutions or looked further for any deeper reasons. Probably you could comment them out and see.

saulipurhonen: For some reason `openCreateFolderModal` property works fine without setter but `openUploadModal` gives warning. I have no better solution for this so I'm happy with this approach.

saulipurhonen:
It seems that modals can't be opened from other than container view? I opened a folder (into objects view) and tried to open either modal but it doesn't work. Seems to be related to that modal components are hosted only in `Containers.vue`

hannyle:
> It seems that modals can't be opened from other than container view? I opened a folder (into objects view) and tried to open either modal but it doesn't work. Seems to be related to that modal components are hosted only in `Containers.vue`

Hmm, this is a good catch 👍 ! I will fix it as well !

swift_browser_ui_frontend/src/views/Containers.vue
hannyle: I figured one of the reasons is that if we use a computed property on `v-model` it needs to have a **setter**.

hannyle:
> It seems that modals can't be opened from other than container view? I opened a folder (into objects view) and tried to open either modal but it doesn't work. Seems to be related to that modal components are hosted only in `Containers.vue`

--> This is fixed!

> I'm not sure if this should be part of this PR but I tried enabling upload button manually and I think it would make sense to close the modal when encryption & upload begins. Now it stays open and a notification is rendered behind the modal background overlay like this:
>  image

--> This is fixed!

> Is it possible to display scroll bar always if content doesn't fit in view?
> 
> I think scroll bar shouldn't be on top of content. Is it possible to move it next to the edge of modal?
> 
> Screenshot 2022-08-19 at 15 33 50

--> This is fixed in Chrome! 
For Firefox, I have v. 103.0.2 and it shows differently from yours. It's a bit trickier to custom the scrollbar on Firefox and I have found no better way within this time limit tbh.

saulipurhonen:
Modal doesn't trap focus and I'm able to navigate out from the modal with keyboard. This might be a problem within the CSC DS modal component itself but maybe we should discuss about it.

hannyle:
> Modal doesn't trap focus and I'm able to navigate out from the modal with keyboard. This might be a problem within the CSC DS modal component itself but maybe we should discuss about it.

yes, this has been a known issue with csc-modal component.

saulipurhonen:
> > Modal doesn't trap focus and I'm able to navigate out from the modal with keyboard. This might be a problem within the CSC DS modal component itself but maybe we should discuss about it.
> 
> yes, this has been a known issue with csc-modal component.

Maybe it'd make sense to open an issue in https://github.com/CSCfi/csc-ui

hannyle:
> In current implementation I'm able to scroll content behind the modal. I think inactive scrolling should be disabled when modal is open.

This is fixed!

saulipurhonen (approved):

saulipurhonen:
Good job, looks good to me :)

swift_browser_ui_frontend/src/common/globalFunctions.js
```diff
@@ -5,4 +5,9 @@ export function toggleCreateFolderModal(folderName) {
...
     store.commit("setFolderName", folderName);
   }
-} 
\ No newline at end of file
+}
+
+export function modifyBrowserPageStyles() {
+  const element = document.getElementById("subContainer");
+  element.classList.toggle("subContainer-additionalStyles");
+}
```
sampsapenna, line 13:
```suggestion
}
```
Missing ending newline, not sure if suggestion actually adds it though.

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -584,13 +591,14 @@ let default_translations = {
         norename:
           "Säiliötä ei voi nimetä uudelleen, " +
           "mutta sen voi kopioida uudella nimellä.",
-        createdFolder: "",
+        createdFolder:
+          "Created folder will be shared with all project members in",
```
sampsapenna, line 595: Missing translation, use e.g.
> Luotu kansio jaetaan kaikille [My research project]:n jäsenille

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -667,13 +678,17 @@ let default_translations = {
...
-        cancel: "Peruuta lataus",
+        cancel: "Peruuta",
         table: {
           name: "Nimi",
           path: "Polku",
           size: "Koko",
           type: "Tyyppi",
         },
+        uploadedFiles:
+          "Uploaded files will be shared with " +
```
sampsapenna, line 689: Suggestion for translation:
> Lähetetyt tiedostot jaetaan kaikille [My research project]:n jäsenille.

swift_browser_ui_frontend/src/common/lang.js
sampsapenna: Alternatively
> Luotu kansio jaetaan kaikille jäsenille projektissa [My research project]

If we don't want to bother with different word ordering in translations.

swift_browser_ui_frontend/src/common/lang.js
sampsapenna: Alternatively:
> Lähetetyt tiedostot jaetaan kaikille jäsenille projektissa [My research project]
if we don't want to bother with different word ordering for translations

swift_browser_ui_frontend/src/components/CreateFolderModal.vue
```diff
@@ -1,58 +1,60 @@
...
+            type="is-primary"
+            :placeholder="$t('message.tagPlaceholder')"
+            :confirm-keys="taginputConfirmKeys"
+            :on-paste-separators="taginputConfirmKeys"
+            data-testid="folder-tag"
+          />
+        </b-field>
+        <p class="info-text is-size-6">
+          {{ $t("message.container_ops.createdFolder") }}
+          <b>{{ $t("message.container_ops.myResearchProject") }}</b>
```
sampsapenna, line 46:
```suggestion
          <b>{{ $store.state.active.name }}</b>
```

swift_browser_ui_frontend/src/components/CreateFolderModal.vue
```diff
@@ -1,58 +1,60 @@
...
+        </p>
+        <c-link
+          :href="`https://my.csc.fi/myProjects/project/${currentProjectID}`"
```
sampsapenna, line 49: The link doesn't work with internal project ID used by Pouta. It will work with the project name from Vuex store, located in `$store.state.active.name` or `this.$store.state.active.name` in the computed variable.

Project name contains ID used by MyCSC, but it's of a long form `project_XXXXXX`. Removing the `project_` part will make it usable in the link.

swift_browser_ui_frontend/src/components/ObjectTable.vue
```diff
@@ -63,7 +63,7 @@
             outlined
             icon-left="upload"
             tag="router-link"
-            :to="{name: 'UploadView', params: {
+            :to="{name: 'UploadModal', params: {
```
sampsapenna, line 66: This shouldn't be a router link to `UploadModal`, which ends up just leading to the front page as this route doesn't exist.
`@click="toggleUploadModal"` works here as well.

swift_browser_ui_frontend/src/components/UploadModal.vue
```diff
@@ -0,0 +1,712 @@
...
+          :no-data-text="$t('message.encrypt.empty')"
+          :pagination.prop="filesPagination"
+          :footer-options.prop="footer"
+        />
+        <p class="info-text is-size-6">
+          {{ $t("message.container_ops.createdFolder") }}
+          <b>{{ $t("message.container_ops.myResearchProject") }}</b>
+        </p>
+        <c-link
+          :href="`https://my.csc.fi/myProjects/project/${currentProjectID}`"
```
sampsapenna, line 60: Same as in container creation modal, use `$store.state.active.name` without the `project_` part to make the link work.

swift_browser_ui_frontend/src/components/UploadModal.vue
```diff
@@ -0,0 +1,712 @@
...
+          class="files-table"
+          :data.prop="dropFiles"
+          :headers.prop="fileHeaders"
+          :no-data-text="$t('message.encrypt.empty')"
+          :pagination.prop="filesPagination"
+          :footer-options.prop="footer"
+        />
+        <p class="info-text is-size-6">
+          {{ $t("message.container_ops.createdFolder") }}
+          <b>{{ $t("message.container_ops.myResearchProject") }}</b>
```
sampsapenna, line 57: Should we use the active project name here as well?

sampsapenna (changes requested):
Wrote suggestions for the missing translations, won't focus on the Finnish more since service team will have differing opinions on wording anyways.

Noticed broken upload button on the object table + some improvement suggestions for the project link and test in the modals.

Overall really nice look and feel to it, LGTM. Will approve after the links are fixed 👍🏼

swift_browser_ui_frontend/src/common/lang.js
hannyle: Thanks 👍

swift_browser_ui_frontend/src/common/lang.js
hannyle: Thanks 👍

swift_browser_ui_frontend/src/components/UploadModal.vue
hannyle: yes :)

sampsapenna (approved):

PR #741, opened 2022-09-06
Use correct URL parameter for API token requests

### LINKED ISSUE #697: Incorrect return value causes token view to break
#### Describe the bug

In the Sharing API access token view (User Information -> tokens),
the view breaks due to tokens API responding with 404 when no tokens are found.

After UI changes to catch errors before API requests finish, this causes the UI to serve
a 404 page, instead of an empty table of tokens with the correct view. Situation can be
fixed by changing the API return value to an empty listing `[]` instead.

#### To Reproduce
1. Go to 'User Information page'
2. Click on 'Tokens'
3. Tokens view fails with 404 error page

#### Expected behavior

Tokens view opens, in which new tokens can be added, and existing ones listed or deleted.

sampsapenna:
### Description

Using the incorrect URL parameter broke the token view. This has now been fixed.

### Related issues

* Fixes #697 

### Type of change

### Changes Made

* Use correct value for current project ID

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/.github/config/.finnishwords.txt
+++ b/.github/config/.finnishwords.txt
@@ -1,4 +1,3 @@
-
 ajaksi
 alapuolisesta
 alla
@@ -282,12 +281,14 @@ ota
 ottaneesi
 ovat
 oy
+päänäkymään
 pääsy
 paina
 painikkeella
 painikkeesta
 päivitä
 pakollista
+palaa
 palautus
 paloitellaan
 paluu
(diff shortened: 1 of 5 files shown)


hannyle:
@sampsapenna Could we have the Changelog updated for this fix ?

hannyle (approved):

PR #749, opened 2022-09-13
Prevent copying folder in place

### LINKED ISSUE #693: Copy folder bug
#### Describe the bug
When copying a folder, it is possible to use same name for copy of the current folder and click "Copy" and nothing happens. 

#### To Reproduce
1. Go to Browser
2. Click on Copy
3. Don't change name for new folder.
4. Click Copy button
5. You are back in Browser page but nothing has happened.

#### Expected behavior
Given that user wants to copy a folder
when user selects Copy 
and then gives a unique folder name
then Copy button is activated
and user clicks button
then user is back at Browser page
and sees copy of the folder in the list.

#### Additional context
Connected to https://github.com/CSCfi/swift-browser-ui/issues/728

sampsapenna:
### Description

`ReplicateContainer` view used old check for container existence before starting a copy, which meant you could copy folder in place overwriting the files and not actually copying them to a new container. Container existence check now rewritten to work with the IndexedDB approach of storing containers, plus it now checks against container specified in the URL.

### Related issues

* Fixes #693 

### Type of change

### Changes Made

* Check copy destination container existence against IndexedDB instead of the old cache
* Check copy destination uniqueness based on the copy source container

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/components/BrowserSecondaryNavbar.vue
+++ b/swift_browser_ui_frontend/src/components/BrowserSecondaryNavbar.vue
@@ -35,8 +35,8 @@
             <c-button
               ghost
               class="navbar-item"
-              @click="copyProjectId"
               data-testid="copy-projectId"
+              @click="copyProjectId"
             >
               <i
                 slot="icon"
--- a/swift_browser_ui_frontend/src/components/FolderOptionsMenu.vue
+++ b/swift_browser_ui_frontend/src/components/FolderOptionsMenu.vue
@@ -111,7 +111,7 @@ export default {
       return this.active.id;
     },
     getContainer: function () {
-      return this.$props.container ? this.$props.container :
+      return this.props.row.name ? this.props.row.name :
         this.$route.params.container;
     },
   },
--- a/swift_browser_ui_frontend/src/views/Replicate.vue
+++ b/swift_browser_ui_frontend/src/views/Replicate.vue
@@ -7,7 +7,7 @@
       v-if="destinationExists"
       type="is-danger"
     >
-      {{ $t('message.replicate.destinationExists') }}
+      {{ $t("message.replicate.destinationExists") }}
     </b-message>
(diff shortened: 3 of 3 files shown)


hannyle:
The copy functionality works for me in general! 👍 

There is one issue is that when the very first folder/container is copied, the copied folder/container does not show in the data table when the route is back to BrowserPage (I guess the containers list are not updated and it could be fixed when I implemented the Copy modal).

Another issue I have is that when I click Copy button in `Replicate.vue`, my console shows this error:

Screenshot 2022-09-15 at 12 16 01

sampsapenna:
Ok, looks like the Copy button error might be related to the container/folder not showing up – might be that I'm dispatching it wrong.

I'll check that out yet before I merge if it's fine 👍🏼

sampsapenna:
> The copy functionality works for me in general! +1
> 
> There is one issue is that when the very first folder/container is copied, the copied folder/container does not show in the data table when the route is back to BrowserPage (I guess the containers list are not updated and it could be fixed when I implemented the Copy modal).
> 
> Another issue I have is that when I click Copy button in `Replicate.vue`, my console shows this error:
> 
> Screenshot 2022-09-15 at 12 16 01

The issue causing container not to show up after starting the copy operation should now be resolved in last commit, together with the error. Thanks for notifying about it 👍🏼

hannyle (approved):
Looks good to me!

P/s: Could the Changelog be updated as well :)

hannyle:
In the new Copy modal, user can also add tags at the same time creating the Copy folder. As I see the function `swiftCopyContainer` does not have params for tags, I guess I could just add another function for saving the tags. Please correct if I'm wrong.

hannyle:
Are we able to copy folder from `Folders shared with you` ? I can see the Copy button there but it doesn't really work.

PR #750, opened 2022-09-19
dependency auto-merge

blankdots:
### Description

dependabot creates a lot of dependency updates, auto-merge them if they have specific labels and all the tests pass

### Type of change

### Changes Made

Added 2 new workflows:
- dependency review which checks dependencies for know vulnerabilities `moderate` or higher
- dependency auto merge, which has 2 steps first one waits for other actions to finish for 30 minutes, and the other merges the PR if all actions are successful. this is done only for PRs with labels `pip dependencies`, `npm dependencies`  and `github actions`, the `docker dependencies` is not added as that sometimes update python or npm versions and that creates issues 

### Testing

- [x] Tests do not apply

### Mentions

From the tests this works about 95% of the time, sometimes it does not recognize the label and a `@dependabot recreate` is required

### DIFF
--- a/.github/workflows/dependency-auto-merge.yml
+++ b/.github/workflows/dependency-auto-merge.yml
@@ -16,13 +16,13 @@ jobs:
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
       - name: Wait other jobs are passed or failed
-        if: ${{ contains(github.event.pull_request.labels.*.name, 'pip dependencies') || contains(github.event.pull_request.labels.*.name, 'github actions') || contains(github.event.pull_request.labels.*.name, 'npm dependencies') }}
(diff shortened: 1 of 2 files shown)


blankdots:
not sure this should be merged if we removed the e2e tests ... as if we have no tests, breaking dependency changes might be merged so we might need #824

csc-felipe:
Perhaps remove auto approving npm packages, but leave others.

blankdots:
> Perhaps remove auto approving npm packages, but leave others.

done

sampsapenna:
> not sure this should be merged if we removed the e2e tests ... as if we have no tests, breaking dependency changes might be merged so we might need #824

e2e tests would also greatly benefit from some changes to the testing workflow, i.e. building an initial state once and performing all the tests without starting from scratch with a new browser session every time. Having to rebuild the IDB for every test is causing a large part of the flakiness, and the rest is from poor API optimization.

We need to fix them before upgrading to Vue 3 though, and I want to write a few new tests for the frontend as well. The eventual upgrade will be a lot simpler if most of the things are part of e2e.

sampsapenna (approved):
LGTM 👍🏼 This seems like a very nice addition

PR #756, opened 2022-09-29
Merge master into devel

csc-felipe:
Let's fix the dissonance between `master` and `develop`!


blankdots (approved):
not sure why the e2e test fail, but this can be merged imo

hannyle:
> not sure why the e2e test fail, but this can be merged imo

e2e test failure can be temporarily ignored due to the amount of API requests from frontend has lead to inconsistent test results in Github actions workflow. I think this PR is good to go.

PR #780, opened 2022-10-17
Make container table options aware of table status

sampsapenna:
### Description

Table options for Object table are aware of current table status, displaying Show/Hide as they are supposed to. This implements the same functionality for the container table view.

### Type of change

### Changes Made

* Mirror `swift_browser_ui_frontend/src/components/ObjectTable.vue` table options functionality in `swift_browser_ui_frontend/src/views/Containers.vue` to make menu aware of current table status.

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/components/ObjectTable.vue
+++ b/swift_browser_ui_frontend/src/components/ObjectTable.vue
@@ -79,6 +79,7 @@
       </div>
 
       <c-menu
+        :key="optionsKey"
         :items.prop="tableOptions"
         options-testid="table-options-selector"
       >
@@ -128,6 +129,7 @@ export default {
       searchQuery: "",
       currentPage: 1,
       checkedRows: [],
+      optionsKey: 1,
       abortController: null,
       filteredObjects: [],
       inCurrentFolder: [],
@@ -593,6 +595,7 @@ export default {
           },
         },
       ];
+      this.optionsKey++;
     },
     setSelectionActionButtons() {
       this.selectionActionButtons = [
--- a/swift_browser_ui_frontend/src/views/Containers.vue
+++ b/swift_browser_ui_frontend/src/views/Containers.vue
@@ -17,6 +17,7 @@
       >
         <SearchBox />
         <c-menu
+          :key="optionsKey"
           :items.prop="tableOptions"
           options-testid="table-options-selector"
         >
@@ -60,6 +61,7 @@ export default {
       currentPage: 1,
       shareModalIsActive: false,
       showTags: true,
+      optionsKey: 1,
       abortController: null,
       containers: { value: [] },
     };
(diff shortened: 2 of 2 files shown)


hannyle:
@sampsapenna Do we have the related issue for this PR ?

sampsapenna:
> @sampsapenna Do we have the related issue for this PR ?
I don't think we have, I just fixed it on the fly after noticing and don't think there's a bug report for it. I can add an issue if you think it's necessary 👍🏼

hannyle:
> > @sampsapenna Do we have the related issue for this PR ?
> 
> I don't think we have, I just fixed it on the fly after noticing and don't think there's a bug report for it. I can add an issue if you think it's necessary 👍🏼

I think that would be nice to have. I think a bug report will help if we need to track back this bug in the future, and other team members could also understand easier :)

hannyle:
> > > @sampsapenna Do we have the related issue for this PR ?
> > 
> > 
> > I don't think we have, I just fixed it on the fly after noticing and don't think there's a bug report for it. I can add an issue if you think it's necessary 👍🏼
> 
> ~I think that would be nice to have~. I think a bug report will help if we need to track back this bug in the future, and other team members could also understand easier :)

@sampsapenna after reviewing this I think this PR wouldn't need a separate bug report.

I don't understand the real purpose of `optionsKey` ? I don't see the difference with or without it ?

sampsapenna:
> > > > @sampsapenna Do we have the related issue for this PR ?
> > > 
> > > 
> > > I don't think we have, I just fixed it on the fly after noticing and don't think there's a bug report for it. I can add an issue if you think it's necessary 👍🏼
> > 
> > 
> > ~I think that would be nice to have~. I think a bug report will help if we need to track back this bug in the future, and other team members could also understand easier :)
> 
> @sampsapenna after reviewing this I think this PR wouldn't need a separate bug report.
> 
> I don't understand the real purpose of `optionsKey` ? I don't see the difference with or without it ?

It's to force re-rendering on data changes, as there were previously issues with the menu if text was changed without a forced update. (text just disappearing and not updating properly) If the key doesn't make a difference then it can go 👍🏼

hannyle:
> > @sampsapenna after reviewing this I think this PR wouldn't need a separate bug report.
> > I don't understand the real purpose of `optionsKey` ? I don't see the difference with or without it ?
> 
> It's to force re-rendering on data changes, as there were previously issues with the menu if text was changed without a forced update. (text just disappearing and not updating properly) If the key doesn't make a difference then it can go 👍🏼

 I don't see the difference or the bug after removing the `optionsKey`, but it might happen in some cases so I think it's good to have it there :)

hannyle (approved):

PR #789, opened 2022-10-28
Update/package versions 10 2022 – combined dependabot upgrades

sampsapenna:
### Description

Aggregated dependabot upgrades, contains all that were opened before 2022-10-28.

### Type of change

### Changes Made

* Update node packages, python packages, docker images
* Deprecate `.value` from database queries, due to library changes

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply


sampsapenna:
> @sampsapenna I think the errors for Spelling Errors and Vue Unit Test could be fixed in this case.

@hannyle Yep, I'll check those out and see what's breaking 👍🏼

blankdots:
btw there is also #750, which might help streamline the dependency update

sampsapenna:
> btw there is also #750, which might help streamline the dependency update

@blankdots I'm aware, it's getting merged after this one 👍🏼 There were some breaking changes in a couple of npm updates, so I decided to merge them this way first. But it seems a real improvement for the future

dockerfiles/Dockerfile-build-crypt
```diff
@@ -87,7 +87,7 @@ RUN export EMCC_CFLAGS="-I/emsdk/upstream/include -L/emsdk/upstream/lib" \
     && emmake make
 
 # Build UI with encryption
-FROM node:16.15.1-alpine3.15 as FRONTEND
+FROM node:19.0.0-alpine3.15 as FRONTEND
```
blankdots, line 90: are you sure you want to go node:19:00 ?

sampsapenna: Good catch, managed to miss that somehow 😅 We most definitely don't, but I'll see if 18 (current LTS) builds

csc-felipe:
Rebasing should solve the python tests, fixed on #824

Issue #743, opened 2022-09-08
Deleting objects doesn't update folder items count


saulipurhonen:
#### Describe the bug

It seems that deleting objects from folder doesn't update IDB and therefore folder items count doesn't change .
Folder size is updated if I manually delete IDB and log in again.

#### To Reproduce
Steps to reproduce the behaviour:
1. Navigate inside folder (objects listing)
2. Delete object(s)
3. Navigate back to folder listing
4. Objects count doesn't change in previously edited folder

#### Expected behavior

Folder objects count should update when there's changes in folders objects count

#### Screenshots

https://user-images.githubusercontent.com/48789543/189100456-903b5cd3-0340-4ffa-be94-1a8341f69ccc.mov

PR #790, opened 2022-11-01
Feature/new footer

rquazi:
### Description

Added the footer with new design for SD-Connect frontend  and also fixed the font of the application. It also fixes the floating footer bug.

### Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- -  Bug Fix https://github.com/CSCfi/swift-browser-ui/issues/814
- -  Bug  Fix https://github.com/CSCfi/swift-browser-ui/issues/859, https://github.com/CSCfi/swift-browser-ui/issues/810, https://github.com/CSCfi/swift-browser-ui/issues/809,

- [x] New feature (non-breaking change which adds functionality)
- - Removed the old footer and added the new footer both in login page and home page https://github.com/CSCfi/swift-browser-ui/issues/694
- - Change the font-family to be Museo-sans https://github.com/CSCfi/swift-browser-ui/issues/719

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [X] Tests do not apply
- [ ] Needs testing (start an issue or do a follow up PR about it)

PR #827, opened 2022-11-14
Fix for updating objects count and size when deleting objects inside folder

hannyle:
### Description

Deleting objects inside a folder should now update the folder's items count and size correctly.

### Related issues
Fixes https://github.com/CSCfi/swift-browser-ui/issues/743

### Type of change

### Changes Made

- Adapted fixes for **IndexedDB flakiness** in branch [release/v2.0.3-large-upload-release](https://github.com/CSCfi/swift-browser-ui/tree/release/v2.0.3-large-upload-release) for `store.js`
- Updated Changelog

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/common/store.js
+++ b/swift_browser_ui_frontend/src/common/store.js
@@ -9,7 +9,6 @@ import {
   getMetadataForSharedContainer,
   getTagsForObjects,
   makeGetObjectsMetaURL,
-  filterSegments,
   tokenize,
 } from "./conv";
 
@@ -248,7 +247,6 @@ const store = new Vuex.Store({
         commit("loading", true);
       }
       let containers;
-      //let sharedContainers;
       let marker = "";
       let newContainers = [];
       do {
@@ -262,7 +260,6 @@ const store = new Vuex.Store({
             cont.tokens = tokenize(cont.name);
             cont.projectID = projectID;
           });
-          await state.db.containers.bulkPut(containers).catch(() => {});
           newContainers = newContainers.concat(containers);
           marker = containers[containers.length - 1].name;
         }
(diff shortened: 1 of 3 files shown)


swift_browser_ui_frontend/src/common/store.js
```diff
@@ -375,7 +385,6 @@ const store = new Vuex.Store({
             obj.containerID = container.id;
             obj.tokens = isSegmentsContainer ? [] : tokenize(obj.name);
           });
-          await state.db.objects.bulkPut(objects).catch(() => {});
```
csc-felipe, line 378: It's nice to get rid of this :)

csc-felipe:
Could you explain what was wrong with the previous logic, and how you solved it?

rquazi:
I was using some bootstrap nav item class which was not responsive on screen resizing. I need to add some other nav class element to make it responsive. Things are not fixed yet :) I am working on it.

csc-felipe:
> I was using some bootstrap nav item class which was not responsive on screen resizing. I need to add some other nav class element to make it responsive. Things are not fixed yet :) I am working on it.

Did you mean to comment on #790?

sampsapenna:
> Could you explain what was wrong with the previous logic, and how you solved it?

@csc-felipe `bulkPut` never replaces the old container or object entries, due to the DB having unique constraints on for `[projectID+name]`. That's how it's supposed to be done, we don't want multiple same entries with the same names, but because of a `.catch(() => {})` added to the end of the `bulkPut` the failure to add containers that already exist wasn't reported in console. If you remove the catch from it you can see how it actually fails for every entry that already had some value in the database.

Because of this existing containers weren't put into the database or updated in any way. Instead, we needed to check if the container already exists in database, and perform an update on the existing container using it's internal key. Replacing existing entries in the database would've worked if the `[projectID+name]` actually was the primary key of the entry, but inbound incremental key was used instead. The `[projectID+name]` index worked to ensure uniqueness though.

The new version actually checks if the old entries are in the database, and just updates them if need be. After giving this a bit more thought removing the `++id` from container and object entries might fix this as well, since it should make the DB use `[projectID+name]` or `[containerID+name]` as primary key.

csc-felipe:
Thanks for explaining! Makes me wonder why did I make the incremental ID as primary key. I might have not realized that we could use `[projectID+name]` or `[containerID+name]`.

Indeed, I'm relieved that `.catch(() => {})` is gone. I felt bad about that, and knew it would cause trouble, but then moved to working in a different project, and couldn't continue.

csc-felipe (approved):

sampsapenna (approved):

hannyle:
Thanks @sampsapenna for helping explain this very comprehensively! 💯

Issue #828, opened 2022-11-14
vueunit.yml failing, check and re-enable when migrating to Vue 3


sampsapenna:
#### Describe the bug

`vueunit.yml` uses `testing-library`, which is no longer supported for Vue 2. This means that, if we are to update to Vue 2.7, we need to deprecate this workflow until move to Vue 3. Migrating to Vue 3 via Vue 2.7 is a simpler alternative, as 2.7 backports a lot of the features provided in Vue 3 – e.g. the composition API as built-in.

#### To Reproduce
Steps to reproduce the behaviour:
```bash
cd swift_browser_ui_frontend
npm i
npm test
```

#### Expected behavior

The jest tests run properly.

#### Additional context

This issue should resolve itself when we migrate to Vue 3, since the library should support it.

PR #829, opened 2022-11-14
Update/package versions 11 2022

sampsapenna:
### Description

Aggregated dependabot upgrades, contains all that were opened before 2022-10-28.

Replaces https://github.com/CSCfi/swift-browser-ui/pull/789 and adds update to `csc-ui` `v0.5.3`, updates to Node.js `v18.12.1` instead of `19.0.0`, and temporarily blocks `vueunit.yml` from workflows ( thus opening issue #828 ).

### Related issues

* Opens #828

### Type of change

### Changes Made

* Update node packages, python packages, docker images
* Deprecate `.value` from database queries, due to library changes
* Remove `vueunit.yml` from workflows for now, since it can't work with Vue 2.7 without major changes ( #828 )

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply
    * Vue unit tests need to be added back, after we update to the next Vue version


blankdots (approved):
LGTM

CHANGELOG.md
```diff
@@ -62,6 +62,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Deprecate `.value` from container and object db listings
+  - `dexie` or `rxjs` removed this from db returns, couldn't find a change causing
+    this in either project, but this is the observed behavior
```
csc-felipe, line 67: It should be a change in `vueuse/rxjs`, and I looked through their changes, and didn't see it either '^^.

csc-felipe (approved):
Changing the `.value` worried me, because when implementing the feature, I noticed that it was the way to access the actual observed object. I would imagine it comes from the breaking changes in [`vueuse` v9.0.0](https://github.com/vueuse/vueuse/releases/tag/v9.0.0), but from their changelog, they do have an example passing `.value` variable, so I'm confused what's going on here.

On the other hand, it seems to work like this.

csc-felipe:
By the way, it's possible to disable the workflow without removing the file (or commenting it out). In the Actions tab in Github, select the workflow on the left, and click the three dots somewhere on the right, which gives the `Disable workflow` option.

sampsapenna:
> By the way, it's possible to disable the workflow without removing the file (or commenting it out). In the Actions tab in Github, select the workflow on the left, and click the three dots somewhere on the right, which gives the `Disable workflow` option.

Ok, thanks for the info 👍🏼 I'll remember this in the future

CHANGELOG.md
sampsapenna: Yep, I searched for any changes in related dependencies for quite long and couldn't see anything hinting towards it. The `.value` just vanished from the returned observers. The best guess from me at this point is that it might have something to do with `composition-api` being part of Vue by default in `2.7`, or some other backported feature.

sampsapenna: I haven't run into any issues with this approach yet, but it troubles me to not actually know the precise reason things have changed

PR #850, opened 2022-11-15
refresh noupload on file entry

sampsapenna:
### Description

Upload using default keys was not working in new modal due to the `refreshNoUpload` method not running on file changes. PR adds this call on change to files that are to be uploaded.

### Type of change

### Changes Made

* Call `refreshNoUpload` on upload modal file input

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - (GH #827) Fixed for updating folder's items count and size when deleting objects inside it
 - (GH #788) Fixed for objects of a copied folder rendering their tags correctly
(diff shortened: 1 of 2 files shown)


hannyle (approved):
LGTM! 👍

hannyle:
@sampsapenna About the Changelog, I thought we would have organised them in chronological order, but I don't think it's applied in our project, for example this PR is listed under (GH #788). What is the guidelines for our Changelog ?

sampsapenna:
@hannyle I mean to add them in chronological order, as in by order of addition. Rebasing sometimes messes up the ordering, as seems to be the case here – I'll need to be more wary of the order in the future 🤔

hannyle:
Thanks for clarifying! I hope the order will be fixed soon 👍

PR #851, opened 2022-11-15
kill upload sessions on finish

sampsapenna:
### Description

Stale upload sessions can lead to the file not reuploaded correctly. This PR fixes that by pruning the old sessions on upload finish.

### Type of change

### Changes Made

* Kill upload session when the upload process finishes

### Testing

- [x] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- (GH #851) Kill upload sessions upon finishing uploads to allow reuploading same files in all cases
 - (GH #884) Fixed multiple bugs
   - Redirected to AllFolders view whenever the selected project changes
   - Modified function and fixed notification for removing a shared permission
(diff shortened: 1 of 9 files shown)


csc-felipe:
> Stale upload sessions can lead to the file not reuploaded correctly. This PR fixes that by pruning the old sessions on upload finish.

I get the impression that there are a few more things going on here. Could you explain the issue and the solution, please? :)

sampsapenna:
> > Stale upload sessions can lead to the file not reuploaded correctly. This PR fixes that by pruning the old sessions on upload finish.
> 
> I get the impression that there are a few more things going on here. Could you explain the issue and the solution, please? :)

@csc-felipe The upload solution doesn't upload same file twice during a single session, which means that the service assumes a file is already uploaded if you try to upload the same file again. This leads to a situation, where the user needs to log out and back in again, if they want to overwrite the files they just uploaded.

This PR changes the behavior so that the files can be immediately re-uploaded once the upload process finishes, but preserves the function to check that the same object name is not used twice during a single upload batch.

There might be a case to check if an object already exists, and ask the user for confirmation. This feature hasn't been requested, however, so this PR only makes it easier to overwrite files during one login session.

blankdots (approved):
lgtm, assuming this gets rebased, feel free to merge

PR #864, opened 2022-11-21
Vault integration

### LINKED ISSUE #602: Add backend support for re-encryption service for adding headers
#### Proposed solution

The header should be pushed to the header storage and re-encryption service instead of Allas. This will allow us to easily rotate the keys when required, and re-encrypt the header with the researcher's public key when they want to download the data back.

#### DoD (Definition of Done)

The upload service can push new headers into the header re-encryption service.

#### Tasks
* [x] add support for vault as default header storage
* [x] add backend functions for header store
* [x] add backend functions for header retrieve
* [x] add backend functions for whitelist entry
* [x] add backend functions for whitelist entry deletion
* [x] add backend functions for public key fetch

#### Testing

Unit tests, integration tests with a mock environment containing a Vault instance with our custom transit plugin.
There's a new issue for tests: #865

csc-felipe:
### Description
Integrating with Vault will allow for more secure handling of encrypted data and encryption keys. The initial implementation requires subsequent changes in the frontend to be fully implemented and usable.

The previous method of storing encrypted headers in Object Storage is preserved as a fallback during the adoption of Vault plugin.

### Related issues
- Closes #602

### Type of change
- [x] New feature
### Changes Made
Implements integration with Vault `c4ghtransit` plugin.

When uploading an object, its header is automatically stored in the Vault backend.

The implementation exposes endpoints so that the frontend can 
- fetch a project's public key
- fetch an object's header
- update a project's whitelisted public key
- delete a project's whitelisted public key

### Testing
- [x] Needs testing - new issue: #865

### DIFF
--- a/swift_browser_ui/common/vault_client.py
+++ b/swift_browser_ui/common/vault_client.py
@@ -1,0 +1,302 @@
+"""Vault c4ghtransit client"""
+import asyncio
+import logging
+import os
+import time
+from base64 import standard_b64encode
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Union
+
+from aiohttp import ClientSession, ClientTimeout
+from aiohttp.web import HTTPError, HTTPGatewayTimeout, HTTPInternalServerError
+from yarl import URL
+
+LOGGER = logging.getLogger("swift_browser_ui.common.vault_client")
+LOGGER.setLevel(os.environ.get("LOG_LEVEL", "INFO"))
+
+
+class VaultServerError(HTTPError):
+    """Service server errors should produce a 502 Bad Gateway response."""
+
+    status_code = 502
+
+
+class VaultClientError(HTTPError):
+    """Service client errors should be raised unmodified."""
+
+    def __init__(
+        self,
+        status_code: int,
+        **kwargs: Any,
+    ) -> None:
+        """Class to raise for http client errors.
+
+        HTTPError doesn't have a setter for status_code, so this allows setting it.
+
+        :param status_code: Set the status code here
+        """
+        self.status_code = status_code
+        HTTPError.__init__(self, **kwargs)
+
+
+@dataclass
+class VaultError:
+    """Unpack a vault error into an object."""
+
+    errors: List[str]
+
+
+def _make_exception(reason: str, status: int) -> HTTPError:
+    """Create a Client or Server exception, according to status code.
+
+    https://developer.hashicorp.com/vault/api-docs#http-status-codes
+
+    :param reason: Error message
+    :param status: HTTP status code
+    :returns VaultServerError or VaultClientError.
+    """
+    if status < 400:
+        LOGGER.error("HTTP status code must be an error code, >400 received %s.", status)
+        return HTTPInternalServerError(
+            reason="Server encountered an unexpected situation."
+        )
+    reason = f"Vault error: {reason}"
+    if status >= 500:
+        return VaultServerError(text=reason, reason=reason)
+    return VaultClientError(text=reason, reason=reason, status_code=status)
+
+
+def _process_error(vault_error: VaultError) -> str:
+    """Return formatted error message.
+
+    https://developer.hashicorp.com/vault/api-docs#error-response
+
+    :param vault_error: Vault error dictionary object
+    :returns: formatted error message as a string.
+    """
+    return vault_error.errors[0]
+
+
+class VaultClient:
+    """AioHttp client that interacts with a Vault server."""
+
+    _vault_token = str()
+    service = "SD-Connect"
+
+    def __init__(self, http_client: ClientSession):
+        """Reuse an existing http client, and load environment variables.
+
+        :param http_client: aiohttp.ClientSession instance
+        """
+        self._client = http_client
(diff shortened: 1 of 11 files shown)


swift_browser_ui/common/vault_client.py
```diff
@@ -0,0 +1,315 @@
...
+                timeout=ClientTimeout(total=10),
+                # headers={
+                #     "X-Vault-Token": await self._token(),
+                #     "X-Vault-Request": "true",
+                #     "Accept": "application/json",
+                #     "Content-Type": "application/json",
+                # },
+            )
+            LOGGER.debug("New HTTP client created")
+        return self._http_client
```
sampsapenna, line 136: Should be possible to use the client saved in `aiohttp.app` when spinning up the server, since the client has a connection pool.

swift_browser_ui/ui/api.py
```diff
@@ -833,3 +833,54 @@ async def get_crypted_upload_session(
...
+    #     + f"&signature={signature['signature']}"
+    #     + f"&valid={signature['valid']}"
+    # )
+
+    return aiohttp.web.Response(
+        status=303,
+        headers={
+            "Location": path,
+        },
+    )
```
sampsapenna, line 886: No need to sign the upload API calls, frontend code can ask the UI to create a pre-signed request for a given path whenever required.

swift_browser_ui/common/vault_client.py
```diff
@@ -0,0 +1,295 @@
+"""Vault c4ghtransit client"""
```
blankdots, line 1: i am wondering if it would have been a good idea to use an already available client instead of implementing our own e.g. `hvac` library seems to do most of the functions below ?

maybe a custom adaptor can be extended: https://hvac.readthedocs.io/en/stable/advanced_usage.html#custom-requests-http-adapter ?

blankdots: welll hvac is not async ... but i am curios why it was not chosen

csc-felipe: One of the main reasons was that it's not async.

But then the API we're accessing is custom-made. And even with `hvac` We'd need very similar code to what we have now. Perhaps the auth would have been simplified, but it doesn't automatically renew tokens, so that logic still had to be implemented. We'd still need the custom error handling, unless I missed something from their codebase.

I think the code is more straightforward this way, too. Nobody has to learn yet another library to make changes to the code, and there's no hidden magic going on.

But if you think that would be an added benefit to using `hvac`, we could re-evaluate, of course.

csc-felipe: Or keep the library updated and the code compatible with it. It just felt like the benefits didn't add up enough in this case.

blankdots: for token renew is: https://github.com/hvac/hvac/blob/v1.0.2/hvac/api/auth_methods/token.py#L348 - maybe that is what you are looking for.

> But if you think that would be an added benefit to using hvac, we could re-evaluate, of course.

no, was just curious of the reasons, I understand them. I am ok with the current solution.

swift_browser_ui/common/vault_client.py
```diff
@@ -0,0 +1,295 @@
...
+    _vault_token = str()
+    service = "SD-Connect"
+
+    def __init__(self, http_client: ClientSession):
+        """Reuse an existing http client, and load environment variables.
+
+        :param http_client: aiohttp.ClientSession instance
+        """
+        self._client = http_client
+        self._base_url = URL(os.environ.get("VAULT_URL", "http://localhost:8200")) / "v1"
```
teemukataja, line 92:
```suggestion
        self._base_url = URL(os.environ.get("VAULT_URL", "http://localhost:8200/v1"))
```
I think `v1` should be part of the configurable address even if it is unlikely to change in the near future

sampsapenna: Seconding this, no need to have the `v1` part hardcoded in the code, even though I don't expect it to change. Other API URLs are configured in a similar manner, like `/v3` in Openstack Keystone URLs.

csc-felipe: ok

swift_browser_ui_frontend/src/entries/main.js
```diff
@@ -309,27 +309,19 @@ new Vue({
...
-          fetch(keyURL)
-            .then(resp => {
-              return resp.text();
-            })
-            .then(resp => {
-              this.$store.commit("appendPubKey", resp);
-            });
-        }
-      }
+      let keyURL = new URL(
```
sampsapenna, line 312: This URL doesn't match the path for getting a public key from the upload runner, thus not resolving to anything but 404. I already wrote a fix for it in the decrypted download implementation, I can open a PR about it on this branch if it's fine by you?

sampsapenna (changes requested):

swift_browser_ui_frontend/src/entries/main.js
csc-felipe: Sure, just push to this branch

sampsapenna (approved):

PR #748, opened 2022-09-13
Feature/edit tags

hannyle:
### Description

Adding modal for editing object's tags. 
Folder's tags and object's tags can now be modified using modal.

### Related issues
Closes https://github.com/CSCfi/swift-browser-ui/issues/724

### Type of change

- [x] New feature (non-breaking change which adds functionality)

### Changes Made

- This PR is rebased on PR [746](https://github.com/CSCfi/swift-browser-ui/pull/746)
- Removed old routes and functions for editing tags and replacing by modal
- Added reusable functions for toggling editing modal
- Updated Changelog

### Testing

- [x] Needs testing (will be tested in PR [746](https://github.com/CSCfi/swift-browser-ui/pull/746) another follow-up PR when the data-table is in stable state)

Issue #724, opened 2022-08-26
Edit tags


ainoc:
Design: https://www.figma.com/file/N0Y3N0nfyMS2wG3gjwZ201/SD-Connect-Prototype?node-id=2023%3A6022

PR #923, opened 2023-01-25
Unify editing tags modal for objects and containers

csc-felipe:
### Description

When making the change to edit tags with a modal, only the tag editor for objects was changed. Editing tags for a container (folder) still used the same editor as the create folder modal. So the design was not the same.

This change makes both container and object tags editor use the same modal, according to the proposed design.

image

### Related issues

- https://github.com/CSCfi/swift-browser-ui/pull/748
- #724

### Type of change

### Changes Made

- Moved the container tag editing functionality from the create folder modal to the edit tag modal

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/components/EditTagsModal.vue
+++ b/swift_browser_ui_frontend/src/components/EditTagsModal.vue
@@ -1,15 +1,14 @@
 <template>
   <c-card class="edit-tags">
     <h4 class="title is-4 has-text-dark">
-      {{ $t('message.objects.editObject') + object.name }}
+      {{ $t('message.editTags') }}
     </h4>
     <c-card-content>
       <b-field
         custom-class="has-text-dark"
-        :label="$t('message.tagName')"
       >
         <b-taginput
-          v-model="object.tags"
+          v-model="tags"
           ellipsis
           maxlength="20"
           has-counter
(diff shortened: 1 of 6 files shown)


swift_browser_ui_frontend/src/components/EditTagsModal.vue
```diff
@@ -123,11 +151,29 @@ export default {
         if (this.$route.name !== "SharedObjects") {
           await this.$store.state.db.objects
             .where(":id").equals(this.object.id)
-            .modify({tags: this.object.tags});
+            .modify({tags: this.tags});
         }
+        // FIXME: Looks like it's missing to save SharedObject tags
```
blankdots, line 156: is this a new issue ? not sure how to interpret this, more details would be nice

csc-felipe: Yes, it might be a bug. I should have removed that one before pushing ^^. Shared objects doesn't use IndexedDB, and without saving it to memory, the UI might not update. But I didn't look into it.

blankdots: ok, then as long as it is created as bug, i am ok.

blankdots (approved):

PR #928, opened 2023-01-27
Allow adding commit hash and url to footer in dev mode

csc-felipe:
### Description

Add new fields to the footer that are only visible in dev mode. The values are extrated with `git` command when building in development mode, e.g. `npm run build-devel`.

image

The version string follows the [`git describe` command](https://git-scm.com/docs/git-describe#_examples).
So in `feature/footer-version | v2.0.0-607-gcaf95d5d`

- `feature/footer-version` is the branch name
- `v2.0.0` indicates the git tag for the release version;
- `607` is the number of commits since the tagged commit;
- `gcaf95d5d` is the shortened commit hash (after removing `g`, so `caf95d5d`)

You can see the commit with `git show caf95d5d`

### Type of change

### Changes Made

- Add a git branch, version and link to the footer, which is only visible in dev mode

### Testing

- [x] Tests do not apply

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - (GH #565) Skip second login button click when first authentication used Haka.
 - (GH #599) Add support for streaming uploads allowing > 1GiB files encrypted uploads
   - Also fixes some bugs related to encrypted upload engine and repeatability
+- Add git commit version and link to footer
 
 ### Changed
 
--- a/dockerfiles/Dockerfile-build-crypt-devel
+++ b/dockerfiles/Dockerfile-build-crypt-devel
@@ -41,7 +41,7 @@ COPY --from=SODIUM /src/libsodium-stable/libsodium-js-sumo/include/ /emsdk/upstr
 COPY --from=SODIUM /src/libsodium-stable/libsodium-js-sumo/lib/ /emsdk/upstream/lib/
 
 ADD https://api.github.com/repos/cscfi/libcrypt4gh/compare/main...HEAD /dev/null
-RUN git clone https://github.com/CSCfi/libcrypt4gh
+RUN git clone --single-branch https://github.com/CSCfi/libcrypt4gh
 
 # We'll skip linking libraries since emcc only produces static libraries
 # Linking sodium at this point causes a linker conflict – thus cutting out $(LIBS)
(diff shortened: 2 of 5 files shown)


blankdots (commented):
i am not sure I am following why we cannot add a command on build or a env var on build to get the latest commit e.g.
```
"build-devel": "APP_VERSION=$(git rev-parse HEAD) && vue-cli-service build --mode development", 
```

or
in vue.config.json
```
process.env.VUE_APP_GIT_HASH = require('child_process').execSync('git rev-parse --short HEAD')
````

or use an npm package.


... and include it directly instead of making it like this and processing the commit in the docker images (which i think we might need to change as part of this PR, if we keep these changes) ?

csc-felipe:
I tried to make it without using `git` command ^^. 

I think the approach you mentioned would not be able to use caching for UI assets in Docker, or in other words, docker might use a cache of a previous commit with the current images. I do think overriding the language files is a good compromise considering how we deploy.

blankdots (changes requested):
modify docker-dev files to make use of this commit

csc-felipe:
The feature is only for deployment, the Dockerfiles in this repo will not be changed.

Or why would you want them to be changed?

blankdots:
> The feature is only for deployment, the Dockerfiles in this repo will not be changed.
> 
> Or why would you want them to be changed?

to me it makes sense to change them as well, if we have dev dockerfiles and we provide them in this repository

csc-felipe:
Yes, but somehow I feel like that would be a different feature. The current feature's intention is to provide the git hash on our deployment environment. Not during development.

blankdots:
> Yes, but somehow I feel like that would be a different feature. The current feature's intention is to provide the git hash on our deployment environment. Not during development.

personally i would find it useful during development as well, as I know which changes are being used, just to spot caching issue. 
That being said, I am fine with not adding it as well.

blankdots (approved):

csc-felipe:
Ok, I'll try your suggestion for making it work both in Docker and local dev.

csc-felipe:
The problem now is that it will always show dirty, as usually the build process modify something. It would be useful, though, to have it there during development.

image

blankdots:
> The problem now is that it will always show dirty, as usually the build process modify something. It would be useful, though, to have it there during development.
> 
> image

that is fine, if that is really troublesome we can remove the string. it is fine as is.

csc-felipe:
Removed the `dirty` flag

blankdots (approved):

PR #952, opened 2023-02-01
Fix footer positioning

csc-felipe:
### Description

The footer positioning was inconsistent across pages. This makes it consistent, and doesn't overlap anymore.

### Type of change

### Changes Made

- Fix footer overlapping other content

### Testing

- [x] Tests do not apply


swift_browser_ui_frontend/src/components/CFooter.vue
```diff
@@ -82,7 +82,9 @@ export default {
 <style lang="scss" scoped>
 @import "@/css/prod.scss";
 
-#c-footer{
+footer{
+  position:relative;
```
blankdots, line 86: this should not be required if the whole page is under display flex :thinking: need to take a look

csc-felipe: Good point, I'll take a look

csc-felipe: Indeed, it doesn't really need that.

blankdots (approved):

PR #963, opened 2023-02-08
Update sphinx and sphinx_rtd_theme

csc-felipe:
### Description

Update Sphinx and the template.

Removed from `requirements.txt`, as it's not needed there.


csc-felipe:
Should fix https://github.com/CSCfi/swift-browser-ui/pull/932

blankdots (approved):

PR #967, opened 2023-02-08
Make footer stick to bottom in large screens on login pages

csc-felipe:
### Description

After https://github.com/CSCfi/swift-browser-ui/pull/952, the footer was not necessarily sticking to the bottom, but only noticeable in larger screens.

This makes the middle content expand as much as possible, pushing the footer to the bottom.

### Type of change

### Changes Made

- Add `flex-grow: 1;` to the main content (`c-row`) in the login pages

### Testing

- [x] Tests do not apply

### DIFF
--- a/swift_browser_ui_frontend/src/pages/IndexOIDCPage.vue
+++ b/swift_browser_ui_frontend/src/pages/IndexOIDCPage.vue
@@ -45,7 +45,8 @@
 <style>
 c-main {
   height: unset;
-  min-height: 100vh
+  min-height: 100vh;
+  justify-content: space-between;
 }
 c-login-card {
   margin: 2rem auto;
--- a/swift_browser_ui_frontend/src/pages/IndexPage.vue
+++ b/swift_browser_ui_frontend/src/pages/IndexPage.vue
@@ -114,7 +114,8 @@ export default {
 <style>
 c-main { 
   height: unset; 
-  min-height: 100vh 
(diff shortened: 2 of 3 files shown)


blankdots:
flex-grow is a way though not sure we might not run into issues with it: 

i think a more flexible approach would be to have the root element as display flex and make use of `height: 100%;` `flex-direction: column;` and  `justify-content: space-between;` or and have the footer `margin-top: auto` see 2nd example of: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Aligning_Items_in_a_Flex_Container#aligning_content_on_the_main_axis or take a look at sd-desktop

csc-felipe:
🤔 I didn't think `flex-grow` could cause issues. I thought it was a standard `flex` way of filling space. But i'll change it to `justify-content: space-between;`

blankdots:
> I didn't think flex-grow could cause issues

i would say it does not necessary cause issues, but to me it seems like we are patching instead of having a proper structure to the page, that would allow consistent position of elements

csc-felipe:
I consider that flex-grow would also give a consistent position of elements ^^. With `space-between`, the space between the main content will also change in addition to the distance to the footer, which to me is less consistent than having only the distance to the footer change. Both approaches are ok to me.

blankdots (approved):

PR #922, opened 2023-01-25
Bugfix/small window modal behaviour

emm1R:
### Description

Copy folder modal does not shrink needlessly for smaller windows.

### Related issues
 "Fixes #910"

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### Testing

- [x] Tests do not apply

PR #974, opened 2023-02-10
Changed private browsing error msg and appearance #922

mradavi:
### Related issues
[Issue: Login page / Replace "﻿Browser not supported" more informative text #922
](https://gitlab.ci.csc.fi/sds-dev/sd-connect/swift-browser-ui/-/issues/922)

### Changes Made

- Changed "Browser not supported" to what's provided on Figma 
- Liisa provided a Finnish translation for it
- Changed the element from <c-login-card-content> to <c-alert> to best reflect the design on figma

### DIFF
--- a/swift_browser_ui_frontend/src/common/lang.js
+++ b/swift_browser_ui_frontend/src/common/lang.js
@@ -46,10 +46,9 @@ let default_translations = {
         inUse: "Bucket name already in use.",
         invalidName: "Bucket name is invalid.",
         createFail: "Bucket creation failed.",
-        idb: "Browser not supported.",
+        idb: "Private mode not supported.",
         idb_text:
-          "This browser is not supported.\n" +
-          "Please, use a supported browser.",
+          "Please switch private mode off to enable login to SD Connect.",    
       },
       dropFiles: "Drag and drop folders here or ",
       help: "Help",
@@ -511,8 +510,9 @@ let default_translations = {
         inUse: "Säiliön nimi on jo käytössä.",
         invalidName: "Säiliön nimi ei kelpaa.",
         createFail: "Säiliön luonti epäonnistui.",
-        idb: "Selain ei tuettu.",
-        idb_text: "Tätä selainta ei tueta. Käytä tuettua selainta.",
+        idb: "Yksityinen selaaminen ei ole tuettu.",
+        idb_text:
+          "Voidaksesi kirjautua vaihda pois yksityisestä selaamisesta.",
       },
       dropFiles: "Vedä ja pudota kansiot tähän tai ",
       help: "Apua",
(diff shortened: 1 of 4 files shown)


blankdots:
please don't link gitlab issues directly

csc-felipe:
Nice first PR!

When there's spelling error check, it's usually because there are new words. You can add them to the dictionary file that lives in `.github/config/`. You could put the word anywhere and use a sorting command, e.g. `sort "${file}" | tr "[:upper:]" "[:lower:]" | sort -u -o "${file}"`.

csc-felipe (changes requested):
sd-connect has a few different "main pages", and you should update the other ones that have the same message as well: `LoginPassword.vue` and `IndexPage.vue`.

To be honest, I'm not very comfortable with the new message, because it's inacurate. The error message may also be displayed in a browser that is not supported, and not only because it's Firefox's private mode.

My suggestion has been to add an entry to user documentation which states the supported browsers, and add a link to it from the error message. But I'm ok with other accurate solutions as well.

cc @ainoc

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -46,10 +46,11 @@ let default_translations = {
         inUse: "Bucket name already in use.",
         invalidName: "Bucket name is invalid.",
         createFail: "Bucket creation failed.",
-        idb: "Browser not supported.",
+        idb: "Firefox in private mode is not supported.",
```
csc-felipe, line 49: It looks like you are not using this string anywhere

csc-felipe (changes requested):
The text string is good now.

However, the template files are not all updated yet.

The changes have to be in all index pages that have it. 
You could search the repository for `message.error.idb` e.g. `grep -r message.error.idb`.
I get
- `LoginPassword.vue`
- `IndexPage.vue`
- `IndexOIDCPage.vue`

So it looks like one of them still need to be updated.

swift_browser_ui_frontend/src/common/lang.js
mradavi: It's used in IndexPage as a card title

csc-felipe:
When you are done with the review comments, you can [dismiss the review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review) and [request a new one](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review).

For choosing who to ask for a review, you can check who has been reviewing and committing recently in the repository.

swift_browser_ui_frontend/src/common/lang.js
csc-felipe: Why is it different in `IndexPage`?

mradavi:
@csc-felipe I didn't update the design of IndexPage since it wasn't included in the issue. The text is updated. 

If the design needs changing there as well, then it would make sense to update the rest of the many error messages on that page as well. IMO warrants a separate issue so that on the design end it could be verified. Perhaps @ainoc would have some input on that?

Current view of IndexPage:
Screenshot 2023-02-13 at 13-15-30 SD Connect

csc-felipe:
Fair enough, sorry about the noise

csc-felipe (approved):

PR #975, opened 2023-02-13
Migrate from npm to pnpm

csc-felipe:
### Description

`npm` installs can sometimes take 10 min, while [`pnpm`](https://pnpm.io) seems to take about a minute tops, on a fresh installation. Otherwise, there doesn't seem to be many changes to the workflow. As a consequence, docker builds will actually be fast now.

Most of the time, the same `npm` command works by just changing it to `pnpm`, except `npx`, which should be `pnpm dlx` or `pnpm exec` I don't know the difference yet.

Dependabot still works, but the vulnerability advisory doesn't seem to: https://github.com/github/roadmap/issues/625

### Type of change

### Changes Made

- Migrate `npm` package manager to `pnpm`

### Mentions
thanks to @blankdots for the suggestion


blankdots (commented):
do we need `vueunit.yml` ?

there are some comments maybe we can remove them ..

csc-felipe:
> do we need `vueunit.yml` ?

It looks like a placeholder for when we have tests

blankdots (approved):

csc-felipe:
I should have put it to merge with the wasm branch, will rebase now that it's merged.

blankdots (approved):

PR #969, opened 2023-02-09
Automate building wasm with npm

csc-felipe:
### Description

Extracted the docker image for building wasm into its own repository, to facilitate reusability and maintainability https://github.com/CSCfi/docker-emscripten-crypt4gh.

This change makes use of it to build wasm files when running `npm` commands.

### Type of change

- [x] New feature (non-breaking change which adds functionality)

### Changes Made

- Remove `Dockerfile-emsdk-deps`
- Update docker files
- Update npm commands
- Update readme

### Testing

- [x] Tests do not apply

PR #976, opened 2023-02-13
Fix building with dockerfiles

csc-felipe:
### Description

There was a mistake in https://github.com/CSCfi/swift-browser-ui/pull/969 that prevented dockerfiles from being able to build, because the node alpine image doesn't seem to have bash installed. This fixes it by not requiring bash.

### Mentions

thanks to @blankdots for spotting it


csc-felipe:
doc tests fail because buefy.org is offline, need to retry later

blankdots (approved):

PR #977, opened 2023-02-13
Allow installing js with --prod

csc-felipe:
### Description

Allow installing js deps with pnpm install --prod for faster and smaller install

The new cypress version takes too long to install, so this fixes it not being installed when not needed.


blankdots (approved):

hannyle:
This PR was closed, but could you explain again how moving dependencies from `devDependencies` to `dependencies` and run with `--prod` flag could help install dependencies faster and smaller ?

blankdots:
> This PR was closed, but could you explain again how moving dependencies from `devDependencies` to `dependencies` and run with `--prod` flag could help install dependencies faster and smaller ?

we noticed that during installation Cypress was installed (which takes a few seconds), even though production does not require cypress, as well another small set of dependencies for testing.

This aims to solve that, installing only the dependencies required to run the software in production omitting the test dependencies.

hannyle:
I understand now. Thank you for explanation! 👍

PR #978, opened 2023-02-13
fix debug logs for vault client

blankdots:
### Type of change

### Changes Made

- fix logs where parameters were missing

### Testing

### Mentions

I would argue that this does not need a Changelog, as the fixes are waay to small, they are mostly misses
The Vue.js Unit Tests was disabled manually so no need to keep it commented.


csc-felipe (approved):
lgtm

PR #979, opened 2023-02-14
Make strings consistent bucket -> folder & remove sd-connect

csc-felipe:
### Description

Update the language file, as it's been inconsistent with the use of `folder` vs `bucket`, and contains branding strings which should not be present.

### Changes Made

- Change text occurrences of `bucket` to `folder`
- Remove sd-connect specific strings
- Updated strings that contain technical wording by removing them

### Mentions

@teemukataja, or someone else, could you please help with the Finnish words? Feel free to push to this branch. Thank you

I didn't check if there are hard coded strings

### DIFF
--- a/swift_browser_ui_frontend/src/common/lang.js
+++ b/swift_browser_ui_frontend/src/common/lang.js
@@ -43,9 +43,9 @@ let default_translations = {
           "performed, contact the service administrator. " +
           "Otherwise head back to the front page from the " +
           "button below.",
-        inUse: "Bucket name already in use.",
-        invalidName: "Bucket name is invalid.",
-        createFail: "Bucket creation failed.",
+        inUse: "Folder name already in use.",
+        invalidName: "Folder name is invalid.",
+        createFail: "Folder creation failed.",
         idb: "Firefox in private mode is not supported.",
         idb_text:
           "Firefox is not supported in private mode. " +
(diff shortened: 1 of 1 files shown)


swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -532,7 +531,7 @@ let default_translations = {
       createFolder: "Luo kansio",
       uploadSecondaryNav: "Lähetä",
       logOut: "Kirjaudu ulos",
-      cscOrg: "CSC – Tieteen Tietotekniikan Keskus Oy",
+      cscOrg: "CSC - Tieteen Tietotekniikan Keskus Oy",
```
blankdots, line 534: that specific character is needed see: https://csc.fi/

swift_browser_ui_frontend/src/common/lang.js
```diff
@@ -77,8 +77,7 @@ let default_translations = {
         sharing_to_one_project: "This folder is shared to one project.",
         sharing_to_many_projects: "This folder is shared to multiple projects.",
         shared_with_read:
-          "This folder is shared with you. You can view files only by" +
-          " using SD Desktop and download files in this folder (Read access)",
+          "This folder is shared with you.",
```
hannyle, line 80: Does this text also related to SD-Connect branding thing so that's why it was removed? 
I found this text from [Figma](https://www.figma.com/proto/N0Y3N0nfyMS2wG3gjwZ201/SD-Connect-Prototype?node-id=2094%3A6510&scaling=min-zoom&page-id=517%3A2532&starting-point-node-id=517%3A2533&show-proto-sidebar=1), but I'm not sure if it's the latest version.

csc-felipe:
Apparently this is not wanted any more.

swift_browser_ui_frontend/src/common/lang.js
hannyle: Actually the term "SD Desktop" is also removed from Figma. I agree that we can recheck later if some texts should be put in here :)

PR #929, opened 2023-01-30
WIP: Decrypt files in frontend when downloading

sampsapenna:
### Description

Add support for decrypting files during downloading, which allows users to easily decrypt the files uploaded into the environment using opaque managed keys. Also contains multiple bugfixes.

### Related issues

* Fixes #704 
* Fixes #705 
* Fixes #779 
* Fixes #819 
* Adds #642 

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update (this needs a follow up PR)

### Changes Made

* Add decryption support to the UI
* Revert back to using separate containers for objects and segments
* Fix some minor issues in Vault integration
* Fix incorrect `struct` usage in WASM
* Fix uploading directly to pseudofolders, both existing and new
* Pre-fill container/folder name when uploading

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply
- [x] Needs testing (start an issue or do a follow up PR about it)

PR #980, opened 2023-02-15
fix url rewrite when path is present

blankdots:
### Description

URL strips path from base url so when deploying we are not getting the correct path

### Type of change

### Changes Made

- concatenate URLs instead of using base on so that path remains

### Testing

- [x] Tests do not apply


hannyle (approved):
Looks good to me! It's nice to know about this issue.

Could we have a Changelog update for this ?

hannyle (approved):

blankdots:
> @blankdots #929 was rebased and had some new changes – meaning this branch needs to be rebased as well. I can do it in case you don't have time for it, but you probably have a better grasp on how to resolve any conflicts in this.

should have merged this before rebasing, but I can try to see what has happened

blankdots:
@sampsapenna apparently with the rebase some formatting issues in C files got fixed

sampsapenna (approved):

PR #982, opened 2023-02-21
Migrate from vue-cli to vite

csc-felipe:
### Description

`vue-cli` is not the recommended tool for `vue` projects any more, and migrating to [`vite`](https://vitejs.dev) is a step towards migrating to `vue3`.

Quite many changes were needed, but it looks like it works as before. Development workflows shouldn't be affected, as the `pnpm` commands are the same.

The most significant change for the project structure is that for a multipage app (with multiple HTML entries), the HTML files need to be declared specifically for each entry.

The biggest difference, as I understood, between both tools is the underlying bundler, and the dev server.

The bundler in `vue-cli` is `webpack`, while for `vite` it's [`rollup`](https://rollupjs.org).

They claim that `vite`'s dev server is superfast, but what it actually does is build dependencies on demand, so it looks faster to start, but while browsing some waiting usually follows. It might anyway be an improvement over `vue-cli`.

The [browser compatibility](https://vitejs.dev/guide/build.html#browser-compatibility) is also supposed to be good enough, even without babel or autoprefixer:
- Chrome >=87
- Firefox >=78
- Safari >=14
- Edge >=88

### Type of change

### Changes Made

- Remove `vue-cli` and dependencies
- Add `vite` and its dependencies
- Migrate `vue-cli` config to `vite` config
- Fix component imports, mostly adding `.vue` to the imports, which also makes type hinting work
- Fix `css` import statements
- Importing `css` styles in `style` blocks was changed to global "import" configured in `vite.config.js`
- Created a middleware plugin for vite to serve the `/browse` page as an SPA, and the other `html` files as well
- Add https configuration for the dev server

### Testing

- [x] Tests do not apply

Please, check that the setup works as expected!


blankdots:
should we wait for #929 before or when we have all the features before the release before we proceed with this ?

csc-felipe:
> should we wait for #929 before or when we have all the features before the release before we proceed with this ?

I think so

hannyle:
Looks good and it works as expected for me 👍

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Migrate from vue-cli to vite
```
hannyle, line 53: As we discussed in a previous daily, it would be nice to include the Github issue number for this change ?

Also, regarding chronological order, could we have the latest change at the bottom (or on top) of the list? I'm okay with either way but just want to have some common rule since some of them are placed in different order now.

csc-felipe: There is no github issue for this, and we had an agreement last month that we would not create new issues on github.

Regarding the order, I agree that it's good to agree on how to do it. I saw someone else starting to write at the top, and did the same.

hannyle: For the Github issue, I mean it could also be the PR number, in this case it is 982, if that makes sense :)

csc-felipe:
Rebased, and added github PR number to the changelog.

csc-felipe:
I'll mark it as WIP while we wait for other PR to be merged

csc-felipe:
I've been developing off this branch, and the dev server doesn't work well with the current configuration.

csc-felipe:
rebased

csc-felipe:
rebased and added https mode for dev server

csc-felipe:
The devserver seems to behave well now. 
The fix was to set the path to javascript file in the `html` files without a leading `.`.
Together with the middleware that serves the correct files.

I still need to test that websockets work, though.

csc-felipe:
websockets work

blankdots (approved):
i personally do not spot anything at this time, was able to build and run it, however best to wait for other reviewers before merging

hannyle:
Worked well for me in most of the pages, some exceptions are for some pages such as `/unauth/`, `/forbid`, `/uidown` which redirected me to `index.html`, I think.

csc-felipe:
@hannyle, it should be fixed now

swift_browser_ui_frontend/vite.config.js
```diff
@@ -0,0 +1,172 @@
...
+  "/upload":              proxyTo,
+  "/enupload":            proxyTo,
+  "/download":            proxyTo,
+  "/sign":                proxyTo,
+  "/replicate":           proxyTo,
+  "/token":               proxyTo,
+  "/ws": {
+    target: `ws${process.env.SWIFT_UI_SECURE_WEBSOCKET}://${process.env.SWIFT_UI_TLS_HOST}:${process.env.SWIFT_UI_TLS_PORT}/ws`,
+    ws: true,
+  },
```
sampsapenna, line 67: What exactly is this meant to proxy? In the old implementation this approach was chosen, to get the websocket URLs to be correct when connecting to Nginx, in this version it just seems to create a circular proxy at first glance. Vite can't connect to the ws when I'm testing this in my own environment, and I can't figure out what I'm missing 😅 

Screenshot from 2023-03-16 12-01-38

csc-felipe: The Vite dev server doesn't have the same kind of configuration as the webpack-dev-server. So I moved the websockets config to the proxy configuration.

https://vitejs.dev/config/server-options.html

sampsapenna: Changed it a bit to allow using the websocket, also fixed the `publicDir` to allow using `libupload.js` and `libupload.wasm` without proxying to backend.

https://github.com/CSCfi/swift-browser-ui/pull/1035/files

csc-felipe: thanks

sampsapenna (approved):
Works nicely, Vite seems considerably quicker in the hot-reloads of files. Took a bit of getting used to the files not getting built initially, though.

Approved, everything seems to work 👌🏼

blankdots (dismissed):

hannyle (approved):

blankdots (approved):

sampsapenna (approved):

Issue #888, opened 2022-12-14
Many elements are not accessible by using Screen reader or VoiceOver


hannyle:
#### Describe the bug

Many elements in SD Connect are not accessible by ScreenReader or VoiceOver. One of reasons is its lacking of `aria-*` attributes in several components.

#### To Reproduce
Steps to reproduce the behaviour:

#### Expected behavior

Elements/components in SD Connect have `aria-*` attributes attached to them.

PR #983, opened 2023-02-22
Make elements more semantic and improve accessibility for screen reader users

hannyle:
### Description

Improving the accessibility for SD-Connect by providing more logical page structure and accessible elements for screen reader users.  
Fixes were made based on rules and guidelines from [WAI-ARIA](https://www.w3.org/WAI/standards-guidelines/aria/)
### Related issues
Fixes https://github.com/CSCfi/swift-browser-ui/issues/888

### Type of change

### Changes Made

- Added `role` and `label` to make HTML elements more semantic 
- Modified `heading` levels 
- Modified BrowserPage to help screen reader to pick up the page in a more logical structure
- Added Finnish translation for `aria-label`
- Updated Changelog

### Testing

- [x] Tests do not apply, but you could have a quick test yourself by installing Chrome extensions. I used **_axe DevTools_** and **_WAVE_**

### Mentions

I couldn't fix for some elements that are nested deeply inside csc-ui components, for example: `<c-button>, <c-data-table>`. The fix should be implemented directly from their source code so either they would fix it or I would create a PR later for `csc-ui`.


csc-felipe (changes requested):
Well done, it's not so easy to follow these guidelines.

It looks to me like the `aria-label` strings should be extracted as translatable strings, and `sd-connect` branding should be left out.

hannyle:
> Well done, it's not so easy to follow these guidelines.
> 
> It looks to me like the `aria-label` strings should be extracted as translatable strings, and `sd-connect` branding should be left out.

It's not 100% ready to be reviewed yet but thanks for your comment :D 💯 
I also just realised that the `aria-label` strings do need to have translation. I will push an update and add reviewers afterwards :)

csc-felipe:
Ah, ok. When it's not ready you can mark it with `WIP: ` or as draft, so it's clear :)

swift_browser_ui_frontend/src/components/ShareModal.vue
```diff
@@ -1,67 +1,74 @@
 <template>
   <c-card class="share-card">
-    <header>
-      <h4 class="title is-4 has-text-dark">
+    <div class="heading">
```
blankdots, line 3: should we use `<c-card-title>` ?

and to encapsulate the `c-button` to use `<c-card-actions>` ?

swift_browser_ui_frontend/src/pages/BrowserPage.vue
```diff
@@ -1,12 +1,16 @@
...
-    <div id="subContainer">
+    <nav>
       <BrowserMainNavbar :langs="langs" />
       <BrowserSecondaryNavbar
         :multiple-projects="multipleProjects"
         :projects="projects"
       />
-      <ProgressBar v-if="isUploading || isChunking" />
+    </nav>
+    <div
```
blankdots, line 10: should this be `<main>` or maybe `c-main` ?

swift_browser_ui_frontend/src/components/ShareModal.vue
hannyle: I also wanted to use `<c-card-title>` in the first place, but it has `<c-title>` underneath and this one has an undesirable underline which we don't have in the current Figma design. 

Using `<c-card-actions>` is a better idea 👍 . I will push the change for it.

swift_browser_ui_frontend/src/pages/BrowserPage.vue
hannyle: The `div` already has `role="main"` which is similar to `<main>`.
https://github.com/CSCfi/swift-browser-ui/blob/bugfix/accessibility-role-label-issues/swift_browser_ui_frontend/src/pages/BrowserPage.vue#L10-L13

`<c-main>` on the other hand I think it's just the name, I don't think they use `<main>` to represent it.

blankdots (approved):

csc-felipe:
@hannyle, you can [dismiss a review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review) that asks for changes after you addressed them, so merging is not blocked any more.

hannyle:
> @hannyle, you can [dismiss a review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review) that asks for changes after you addressed them, so merging is not blocked any more.

Should I dismiss the review when I was actually waiting for the approval after I made the fix for those changes requested ?

csc-felipe:
> Should I dismiss the review when I was actually waiting for the approval after I made the fix for those changes requested ?

Yes, I think that's the way. So it's clear you're done with the comments.

hannyle:
> > Should I dismiss the review when I was actually waiting for the approval after I made the fix for those changes requested ?
> 
> Yes, I think that's the way. So it's clear you're done with the comments.

Ok, I usually just re-request for review but good to know that I could also do this from now on. Thanks! :)

csc-felipe (approved):
Looks great

hannyle:
Rebased on `develop` and it said dismissed stale reviews

blankdots (approved):

PR #984, opened 2023-02-22
Migrate to using pyproject.toml with hatch as build tool

csc-felipe:
### Description

Using `pyproject.toml` seems to be the recommended way for building and installing python projects.

Nothing else needs to change, the project still installs as usual with `pip`.

The build system is now [hatch](https://hatch.pypa.io/latest/), as it was one of the [recommended tools](https://packaging.python.org/en/latest/tutorials/packaging-projects/), and is an [official project](https://github.com/pypa/hatch). It has quite many features, but we don't even need to install it.

### Type of change

### Changes Made

- Migrate from `setup.py` to `pyproject.toml`
- Added the frontend `dist` folder, otherwise the build tries to resolve it as a local directory, and fails to install when it doesn't exist
- Update `cryptography`, as dependency review flagged the current version as vulnerable

### Testing

- [x] Tests do not apply

### Mentions

I was wondering if we should remove the requirements file, as it is not necessarily needed. Any opinions?


blankdots (approved):
i think this could can be merged, as it does not affect that much the current development

PR #987, opened 2023-02-24
WIP: Add pre-commit hooks

csc-felipe:
### Description

Add pre-commit hooks and checks.

Also added a github action. If it works, and others agree, we could change a bit the github workflows and tox configuration.

Update your environment and enable the git hooks by running:

    pip install -Ue .[docs,test,dev]
    pre-commit install

### Type of change

### Changes Made

- Added pre-commit configuration
- fixed raised issues from the checks
- moved `pyspelling` configuration to root folder
- use [`ruff`](https://beta.ruff.rs/) instead of `flake8` and `isort`

### Testing

- [x] Tests do not apply

### Mentions

It might be easier to review by looking at commits individually.

I tried to use the `pyupgrade` fixes from `ruff`, but that broke the tests, and I didn't want to investigate. There are a few other tests that might be interesting to consider in the future: e.g. [`tryceratops`](https://beta.ruff.rs/docs/rules/#tryceratops-try), which checks for how to handle exceptions.


csc-felipe:
Interesting that the pre-commit action fails because it sorts differently than my system's sort.

In the action sort puts **ä** after the whole alphabet, while in my system it is considered the same as **a**.

csc-felipe:
Rebased, and added `export LC_ALL=C` to make sorting consistent across machines, as suggested by sort's manual:

       *** WARNING *** The locale specified by the environment affects
       sort order.  Set LC_ALL=C to get the traditional sort order that
       uses native byte values.

csc-felipe:
Skip eslint pre-commit run when it is not installed e.g. in the pre-commit github action.

blankdots (commented):
:thinking: i think we should wait on this, this seems to touch a lot of ongoing files. I would rather see functionality added first and standardization last, for the simple reason of that there is too much ongoing work ...

csc-felipe:
You're right. I'll keep it as WIP while we wait

PR #989, opened 2023-02-28
Make selected display options consistent

hannyle:
### Description

User's selected Display Options are now saved in IDB so they are consistent when browsing between different pages/components

### Related issues
Fixes Gitlab issue 932

### Type of change

### Changes Made

- Saved Display Options as an object to `container` and `object` in IDB

### Testing

- [x] Tests do not apply

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - (GH #965) Show notification when changing permission in share modal
 - (GH #969) Automate building wasm with npm
 - (GH #983) Make elements more semantic and improve accessibility for screen reader users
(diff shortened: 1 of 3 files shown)


hannyle:
> This works well, cool!
> 
> I was just wondering, though, that if a person turns off tags or pagination for one folder, shouldn't it disable for all folders instead of per folder? I wonder what would be most common, that person never wants to see tags of pagination, or that sometimes they may want to toggle it only for a certain folder 🤔

That's a good question. 
I think the answer should come from users through usability tests or real user's feedback. The UX team did some usability tests with our new UI some time ago but I'm not sure if they tested this feature. We could discuss this with @ainoc :)

I personally think we could have both options :D

csc-felipe:
Let's see if it comes up at some point or not

csc-felipe (approved):

PR #1001, opened 2023-03-02
Add workflow for updating pnpm packages

csc-felipe:
### Description

Dependabot doesn't update `pnpm` packages, so here's an action to create a PR with `pnpm` updates.

It's not perfect, though.
1. It runs when a new PR is created, as cron jobs only run on the default branch
2. It uses `GITHUB_TOKEN`, so the new PR doesn't trigger other workflows
3. PR has to be closed and opened again to run checks

More info:
- https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow
- https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs
- https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md#use-case-create-a-pull-request-to-update-x-periodically

### Type of change

### Changes Made

- Added a new github workflow for automatically updating pnpm packages
- Change workflows from `push` to `pull_request`, so that checks run in the automated PR

### Testing

- [x] Tests do not apply

### DIFF
--- a/.github/workflows/pnpm-autoupdate.yml
+++ b/.github/workflows/pnpm-autoupdate.yml
@@ -1,0 +1,44 @@
+name: PNPM auto update packages
+
+on: 
+  pull_request:
+    branches: [devel]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: ['18']
+    name: pnpm packages update
+    steps:
+      - uses: actions/checkout@v3
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 7
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'pnpm'
+          cache-dependency-path: swift_browser_ui_frontend/pnpm-lock.yaml
+      - name: Install dependencies
+        run: |
+          pushd swift_browser_ui_frontend
+          pnpm update
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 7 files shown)


csc-felipe:
It seems to work: https://github.com/CSCfi/swift-browser-ui/pull/1002

.github/workflows/pnpm-autoupdate.yml
```diff
@@ -0,0 +1,41 @@
...
+          pushd swift_browser_ui_frontend
+          pnpm update
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Update pnpm packages
+          title: PNPM package updates
+          body: Automated PR with PNPM updates
+          branch: update/pnpm-updates
```
blankdots, line 40: maybe we can create a new branch, so adding a random nb at the end would do.
They will get deleted on merge either way.

csc-felipe: If we do that, we may end up with multiple pull requests. When the packages are outdated, each new pull request would create a new pnpm update pull request. If we use the same branch name (current version), we get only one PR.

More details in the github action documentation that creates PRs https://github.com/peter-evans/create-pull-request#action-behaviour

blankdots (dismissed):

csc-felipe:
Not all the checks in that PR are running 🤔

csc-felipe:
I'm thinking that those checks might need to have `on: pull_request` as well.

blankdots (approved):

csc-felipe:
It seems to work if the workflows have `on: pull_request`. But the generated PR will only work after this is merged to `devel`, as it changes `push` to `pull_request` in workflows.

blankdots (approved):

PR #1006, opened 2023-03-06
migrate away from buefy

csc-felipe:
### Description

All `buefy` components have been removed from `devel`, and this PR merely removes the `buefy` package and its references.

### Changes Made

- Remove `buefy` dependency

PR #1007, opened 2023-03-06
Replace Buefy Taginput and Taglist component

hannyle:
### Description

Replacing Buefy Taginput and Taglist component by our own component based on  [csc-ui Tag](https://csc-design-system-docs.web.app/c-tag)

### Related issues
GL issue 944

### Type of change

### Changes Made

- Created TagInput component to replace Buefy `<b-taginput>` component and be reusable in other components/views
- Removed codes no longer needed
- Updated Changelog

### Testing

- [x] Tests do not apply

### Mentions

As part of PR https://github.com/CSCfi/swift-browser-ui/pull/1006
I didn't  make changes for `SharedDetails` component since it is removed


blankdots (approved):

csc-felipe (dismissed):
This is really nice!

Just a couple of minor comments:
- It would be nice if the taginput would have an on-focus border like the other inputs
- It would also be nice if the keys that trigger a tag to be created would continue to be the same
- I noticed there was still a `b-taglist` in `components/SharedDetails.vue`. Could it be changed as well?

hannyle:
> Just a couple of minor comments:
> 
> * It would be nice if the taginput would have an on-focus border like the other inputs

--> This is now fixed!

> * It would also be nice if the keys that trigger a tag to be created would continue to be the same

--> I brought back all the Keys that we previously had :)

> * I noticed there was still a `b-taglist` in `components/SharedDetails.vue`. Could it be changed as well?

--> I explained this in the **Mentions**, the `SharedDetails.vue` will be removed in the end so I don't think it's necessary to update this component

csc-felipe:
> --> I explained this in the Mentions, the SharedDetails.vue will be removed in the end so I don't think it's necessary to update this component

Sorry I missed that

swift_browser_ui_frontend/src/components/ShareModal.vue
```diff
@@ -423,15 +431,20 @@ export default {
...
 
-  .field {
-    margin: 2rem 0 0 0;
+  #shareid-tags {
+    margin: 1rem 0;
+    border: 1px solid $csc-dark-grey;
+  }
+
+  #shareid-tags:focus-within {
+    border: 2px solid $csc-primary;
```
csc-felipe, line 447: Is this one also needed if the component already has the same styling?

swift_browser_ui_frontend/src/components/ShareModal.vue
```diff
@@ -423,15 +431,20 @@ export default {
...
 
   c-link > span {
     font-size: 0.875rem;
   }
 
-  .field {
-    margin: 2rem 0 0 0;
+  #shareid-tags {
+    margin: 1rem 0;
+    border: 1px solid $csc-dark-grey;
```
csc-felipe, line 443: Shouldn't this be part of the taginput component?

hannyle: I would love to. But one thing is the `<c-select>` component has a darker border and current it cannot be modified from our repo, so I had to make the Taginput in this ShareModal a bit darker color to match with the `c-select`

swift_browser_ui_frontend/src/components/ShareModal.vue
hannyle: This relates to my other comment. Since I already put the border for Taginput in this ShareModal, I need to specify the other border when it is focused.

swift_browser_ui_frontend/src/components/ShareModal.vue
csc-felipe: I see what you mean! Makes me wonder why aren't all the fields looking the same, as they come from `csc-ui`

csc-felipe (dismissed):
lgtm

hannyle:
TagInput's border color is set to dark color by default now.

csc-felipe (approved):

PR #1021, opened 2023-03-09
Bugfix/refactor pnpm update

blankdots:
### Type of change

### Changes Made

- same style for YAML files
- refine when to run pnpm updates when dependabot update happen for npm or yarn ecosystem, as well allow github actions to run when pushing to `dependabot/**` branches

### Testing

- [x] Tests do not apply

### Mentions

refactored a bit as new understanding how this runs was observed

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -7,7 +7,7 @@ on:
     # The branches below must be a subset of the branches above
     branches: [master, devel]
   schedule:
-    - cron: '0 17 * * 2'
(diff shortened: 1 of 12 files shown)


csc-felipe (approved):
As I understand, adding the dependabot push branch will not make a push that happens inside a workflow to trigger other workflows, but it doesn't hurt to try.

blankdots:
> As I understand, adding the dependabot push branch will not make a push that happens inside a workflow to trigger other workflows, but it doesn't hurt to try.

i see, well let us try, if not I will try something else

PR #1023, opened 2023-03-10
Feature/replace buefy dialog

emm1R:
### Description

Replace buefy.dialogs with csc-ui modals

### Related issues
GL 944


csc-felipe:
You put the target branch as `master`, but it should be `devel`, I'll change that.

csc-felipe:
Can we close this, as you opened another PR?

emm1R:
I think I did this by accident. Oops

PR #1030, opened 2023-03-13
Replace b-upload with a new component CUploadButton

csc-felipe:
### Description

Part of migrating away from `buefy`, replaces `b-upload` with a new component.

I like this approach because it uses `c-button`, but when the file picker opens, it doesn't have focus, which is a bit weird.
### Type of change

### Changes Made

- Create a new upload button component: `CUploadButton`
- Replace `b-upload` with `CUploadButton`

### DIFF
--- a/swift_browser_ui_frontend/src/components/CUploadButton.vue
+++ b/swift_browser_ui_frontend/src/components/CUploadButton.vue
@@ -1,0 +1,70 @@
+Inspired by https://github.com/buefy/buefy/blob/3b3ae60e448ddfd669f20570d40812fd1e041473/src/components/upload/Upload.vue
+
+<template>
+  <div class="upload-btn-wrapper">
+    <c-button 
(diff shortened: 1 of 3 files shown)


swift_browser_ui_frontend/src/components/CUploadButton.vue
```diff
@@ -0,0 +1,70 @@
...
+      if (value.length === 0) {
+        if (!this.newValue) return;
+      } else {
+        let newValues = false;
+        for (let i = 0; i < value.length; i++) {
+          const file = value[i];
+          this.newValue.push(file);
+          newValues = true;
+        }
+        if (!newValues) return;
```
hannyle, line 56: Looks great to me!
Just curious what do we need this `newValues` for?

csc-felipe: So it doesn't emit an event

PR #1031, opened 2023-03-13
make dockerfiles build pyproject.toml

blankdots:
### Description

Using `hatch` as a build tool seems to have issues with finding the symbolic link of the static folder (and seems poetry has similar issues). Switching to setuptools seems to create more issues, so trying the next best thing which is remove the symbolic link from dockerfiles.

This was introduced in: https://github.com/CSCfi/swift-browser-ui/pull/984

### Type of change

### Changes Made

- small additions in `pyproject.toml`
- remove `static` folder from dockefiles so that build can happen normally

### Testing

- [x] Tests do not apply

### DIFF
--- a/dockerfiles/Dockerfile-request
+++ b/dockerfiles/Dockerfile-request
@@ -4,6 +4,10 @@ COPY pyproject.toml /root/swift_request/pyproject.toml
 COPY README.md /root/swift_request/README.md
 COPY swift_browser_ui /root/swift_request/swift_browser_ui
 
+# remove this link as it creates issues with hatch
+# in pyproject.toml
+RUN rm /root/swift_request/swift_browser_ui/ui/static
+
 RUN --mount=type=cache,target=/root/.cache/pip \
     pip install --upgrade pip \
     && pip install /root/swift_request
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,7 @@ name = "swift-browser-ui"
 dynamic = ["version"]
 description = "Object browser Web UI for Openstack Swift API"
 readme = "README.md"
+requires-python = ">=3.10"
 license = "MIT"
 authors = [
     { name = "CSC Developers" },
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -67,6 +68,7 @@ swift-x-account-sharing = "swift_browser_ui.launcher:run_sharing"
 
 [project.urls]
 Source = "https://github.com/CSCfi/swift-browser-ui"
+Documentation = "https://swift-browser-ui.readthedocs.io"
```
csc-felipe, line 71: I didn't know about this ^^

dockerfiles/Dockerfile-request
```diff
@@ -4,6 +4,10 @@ COPY pyproject.toml /root/swift_request/pyproject.toml
 COPY README.md /root/swift_request/README.md
 COPY swift_browser_ui /root/swift_request/swift_browser_ui
 
+# remove this folder as it creates issues with hatch
+# in pyproject.toml
+RUN rm -rf /root/swift_sharing/swift_request/ui/static
```
csc-felipe, line 9: nitpick: they are links, so `rm` should be enough

dockerfiles/Dockerfile-request
```diff
@@ -4,6 +4,10 @@ COPY pyproject.toml /root/swift_request/pyproject.toml
 COPY README.md /root/swift_request/README.md
 COPY swift_browser_ui /root/swift_request/swift_browser_ui
 
+# remove this folder as it creates issues with hatch
```
csc-felipe, line 7: Perhaps the comments should say that it's a link? It might be confusing to someone else who might read this later otherwise

pyproject.toml
blankdots: these get build from the docs folder on readthedocs, haven't been updated ... but it is the official external documentation link

csc-felipe (approved):

PR #1034, opened 2023-03-16
Migrate to vue3

csc-felipe:
### Description

Migrate vue2 → vue3.

Most of the breaking changes don't apply to our code, so it's not such a big code change. But still quite a large PR.

https://v3-migration.vuejs.org/breaking-changes/#template-directives
https://v3-migration.vuejs.org/breaking-changes/v-model.html
https://vuejs.org/guide/reusability/custom-directives.html#directive-hooks

### Related issues

GL 943

~needs #982 #1006~

### Type of change

### Changes Made

- Update libraries
  - vue
  - vuex
  - vue-i18n
  - vue-router
   - fix changing project
- refactor index page app creation to be less repetitive
- update csc-ui vue directive to work with vue3
- add IndexedDB workaround for https://github.com/dexie/Dexie.js/issues/613
- remove `LAST_ACTIVE` cookie reference, as it's not being set anywhere
- fix bindings in autocomplete fields by using `:query` instead of `v-model`
- some values that were not reactive, now are, and can't be passed as is to Dexie, they need `toRaw()` from `vue`

### Testing
- [x] Tests do not apply
- [x] Manual tests are in order to confirm everything works as before

### Mentions

In VSCode, the Vue extension change from Vetur to [Volar](https://marketplace.visualstudio.com/items?itemName=Vue.volar). So you might want to update your setup.

Also, remember to `pnpm install` to update the libraries.

Vue development browser extension now works


swift_browser_ui_frontend/src/entries/index_app_factory.js
```diff
@@ -0,0 +1,40 @@
...
+                          i18n.locale +
+                          "; path=/; expires="
+                          + expiryDate.toUTCString();
+      },  
+    },
+    ...Component,
+  })
+    .use(i18n)
+    .directive("csc-model", vControl);
+}
```
hannyle, line 40: I followed the `csc-ui` instruction and got it worked this way:
- replacing `csc-model` to `vControl` from `index_app_factory.js`
- adding `v-control` and replacing `v-csc-model` to `v-model` like the example
<br/>

Screenshot 2023-03-17 at 12 29 30

csc-felipe: Well, that sounds easy ^^

thank you

csc-felipe:
Ready for review

csc-felipe:
rebased

swift_browser_ui_frontend/src/entries/main.js
```diff
@@ -268,55 +243,46 @@ new Vue({
...
+            projects.indexOf(
+              projects.find(e => e.id == this.$route.params.project),
+            )
+          ];
       }
       this.$store.commit("setActive", active);
 
       if (document.location.pathname == "/browse") {
-        this.$router.push({
+        this.$router.replace({
```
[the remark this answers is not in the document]
csc-felipe: So the user can't go back to `/browse`, which is a page that doesn't really exist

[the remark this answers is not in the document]
hannyle: I see

swift_browser_ui_frontend/src/common/router.js
```diff
@@ -1,15 +1,17 @@
...
-export default new Router({
-  mode: "history",
+export default createRouter({
+  history: createWebHistory(),
   routes: [
+    {
+      path: "/browse",
+      name: "AllFolders",
+      component: FoldersView,
+    },
```
hannyle, line 14: Do we really need this path? It's also related to my previous comment about router.replace 
I never get to this `/browse` actually :thinking:

hannyle:
I'm getting these warnings

Screenshot 2023-03-21 at 14 19 21

csc-felipe:
I see the warnings too. The `vue-router` warning is what happens during component initialization. The other warnings are from `i18n` vite plugin. It's not recommended to have HTML in the strings, and it was a deliberate choice, as I didn't know at the time (and still don't) how else to accomplish the same thing. But feel free to create a ticket for improving that.

swift_browser_ui_frontend/src/common/router.js
csc-felipe: Do you think it should be removed?

hannyle:
> I see the warnings too. The `vue-router` warning is what happens during component initialization. The other warnings are from `i18n` vite plugin. It's not recommended to have HTML in the strings, and it was a deliberate choice, as I didn't know at the time (and still don't) how else to accomplish the same thing.that.

I could also help take a look if I can fix this, if you don't mind.

swift_browser_ui_frontend/src/common/router.js
hannyle: When I tested from my side without this part, I think it didn't affect anything so probably it could be removed.

hannyle:
After the fix for `autocomplete`, the previous issue has gone but I think in the Upload modal, when you type some letters it doesn't suggest the result anymore. It still works fine for Searchbox I think.

csc-felipe:
> After the fix for `autocomplete`, the previous issue has gone but I think in the Upload modal, when you type some letters it doesn't suggest the result anymore. It still works fine for Searchbox I think.

It should be fixed now

blankdots (commented):
from my point of view stuff works, and I cannot spot anything in the code.

All i can spot in the build are these warnings:
```
emcc: warning: warnings in JS library compilation [-Wjs-compiler]
emcc: warning: running limited binaryen optimizations because DWARF info requested (or indirectly required) [-Wlimited-postlink-optimizations]
vite v4.2.0 building for production...
✓ 490 modules transformed.
rendering chunks (60)...warnings when minifying css:
▲ [WARNING] Unexpected ";" [css-syntax-error]

    <stdin>:10223:1:
      10223 │ };
            ╵  ^
```

blankdots:
about the css warning @hannyle pointed out these lines: https://github.com/CSCfi/swift-browser-ui/blob/feature/vue3/swift_browser_ui_frontend/src/components/SearchResultItem.vue#L106-L109

blankdots (dismissed):
from my point of view it works, but please wait for other reviewers

hannyle:
I have this issue when using the Edit tags modal, the tags were not properly saved and might be a lingering issue of dexie :/

Screenshot 2023-03-24 at 12 01 51

sampsapenna (commented):
I'm experiencing a bug, where after login the navbar is partially empty because the username wasn't in the URL from the beginning, maybe it's breaking reactivity? After refreshing everything is properly rendered, but if I manually navigate to `/browse` again, the bug is repeated.

It might have something to do with router being more strict about the router parameters, complaining that the initial `/browse` doesn't have those defined while some component tries accessing them.

Screenshot from 2023-03-24 14-29-37

Screenshot from 2023-03-24 14-30-46

sampsapenna:
> I noticed that bug as well, I think @csc-felipe created a separate issue for it? https://gitlab.ci.csc.fi/sds-dev/sd-connect/swift-browser-ui/-/issues/972

Seems like it, thanks for the info 👍🏼

sampsapenna:
> I noticed that bug as well, I think @csc-felipe created a separate issue for it? https://gitlab.ci.csc.fi/sds-dev/sd-connect/swift-browser-ui/-/issues/972

Seems like it, thanks for the info 👍🏼

sampsapenna:
> I noticed that bug as well, I think @csc-felipe created a separate issue for it? https://gitlab.ci.csc.fi/sds-dev/sd-connect/swift-browser-ui/-/issues/972

Seems like it, thanks for the info 👍🏼

sampsapenna (dismissed):

csc-felipe:
> I have this issue when using the Edit tags modal, the tags were not properly saved and might be a lingering issue of dexie :/

These issues are because of the reactivity model in Vue3, and it wraps _everything_ in a `ref`.

csc-felipe:
Makes me think that we may not even need the workaround... But well, it's a good workaround anyway

csc-felipe:
What it means is that the array with tags, for example, becomes a `ProxyArray`, actually an object. So Dexie doesn't like it ^^.

hannyle:
I wonder why it happens only with Edit tags but not with creating folder or copy folder?

csc-felipe:
It happens only when the Proxy object is passed to Dexie, and Dexie tries to do something with it. And I exagerated that _everything_ becomes a reactive proxy object. This same case probably doesn't apply to the previous issue.

csc-felipe:
That's hopefully fixed now!

hannyle:
> That's hopefully fixed now!

Yes it's fixed 👍

hannyle:
There is a similar issue with `Display Options` in ObjectsView that changing 3 options shows the same error with Dexie.

I did a quick research and understand that because Vue 3 has built-in `reactivity` , we could use `liveQuery()` to have reactive data-fetching or query, for example:

Screenshot 2023-03-28 at 9 03 30

csc-felipe:
I'm not sure I understand the suggestion to use `liveQuery`, because it's usually for reading data, but that db call is for writing data.

hannyle:
Well that's true, hm.. :thinking:

csc-felipe:
`toRaw` also works here, so we don't have to pass the reactive values to dexie.

hannyle (approved):
Looks good to me! Great job! 🎉

PR #1036, opened 2023-03-21
serve correct index oidc

blankdots:
### Description

When deploying with OIDC_ENABLED the `index.html` page seems to be missing 
```
/app $ ls -al /usr/local/lib/python3.10/site-packages/swift_browser_ui/ui/static/
total 3124
drwxr-xr-x    3 root     root          4096 Mar 20 13:58 .
drwxr-xr-x    4 root     root          4096 Mar 20 13:58 ..
-rw-r--r--    1 root     root             0 Mar 20 13:58 .gitkeep
-rw-r--r--    1 root     root          1419 Mar 20 13:58 400.html
-rw-r--r--    1 root     root          1415 Mar 20 13:58 401.html
-rw-r--r--    1 root     root          1415 Mar 20 13:58 403.html
-rw-r--r--    1 root     root          1417 Mar 20 13:58 404.html
-rw-r--r--    1 root     root          1415 Mar 20 13:58 503.html
drwxr-xr-x    2 root     root          4096 Mar 20 13:58 assets
-rw-r--r--    1 root     root          1326 Mar 20 13:58 browse.html
-rw-r--r--    1 root     root          1330 Mar 20 13:58 index_oidc.html
-rw-r--r--    1 root     root        212948 Mar 20 13:56 libupload.js
-rwxr-xr-x    1 root     root       2926364 Mar 20 13:56 libupload.wasm
-rw-r--r--    1 root     root           993 Mar 20 13:58 login.html
-rw-r--r--    1 root     root          1419 Mar 20 13:58 login2step.html
-rw-r--r--    1 root     root          1251 Mar 20 13:58 loginpassword.html
-rw-r--r--    1 root     root           995 Mar 20 13:58 select.html
```
This is an attempt to fix thisby using https://vitejs.dev/guide/api-plugin.html#transformindexhtml

### Type of change

### Changes Made

- if oidc is enabled replace the `index.js` with `index_oidc.js`
- removed the `index_oidc.html` page as I don't think it is required with this replacement

### Testing

### Mentions

**I am open to any other fixes**


csc-felipe:
Also, the `login2step.html` file points to `index.js` as well. Doesn't it get replaced? You might need to check that it replaces only on `index.html`

csc-felipe:
So when OIDC is enabled, a new step is added before the existing step. So the existing step becomes the second, and the OIDC becomes the first. And we have this trick to make it happen.

blankdots:
> This looks good. I just don't see the check for if oidc is enabled.

done

blankdots:
> So when OIDC is enabled, a new step is added before the existing step. So the existing step becomes the second, and the OIDC becomes the first. And we have this trick to make it happen. The backend forwards to correct files.

ok, not sure what to do with that information, do you have any other fix that can be applied all is fine from my side

csc-felipe:
> ok, not sure what to do with that information, do you have any other fix that can be applied all is fine from my side

No, it was partly talking to myself

swift_browser_ui_frontend/vite.config.js
```diff
@@ -101,6 +99,20 @@ const multipagePlugin = () => ({
   },
 });
 
+const htmlPlugin = (oidc) => {
+  if (oidc) {
+    return {
+      name: "html-transform",
+      transformIndexHtml: {
+        enforce: "pre",
+        apply: 'build',
```
csc-felipe, line 108: I think we don't want this `apply`, so that it will do the same also on the development server.

csc-felipe (approved):

PR #1040, opened 2023-03-23
Add pre-commit hooks and github workflow

csc-felipe:
### Description

Add pre-commit hooks and checks.

Also added a github action. If it works, and others agree, we could change a bit the github workflows and tox configuration.

Update your environment and enable the git hooks by running:

    pip install -Ue .[docs,test,dev]
    pre-commit install

### Related issues

- follow-up from #987

### Type of change

### Changes Made

- Added pre-commit configuration
- fixed raised issues from the checks
- moved `pyspelling` configuration to root folder
- use [`ruff`](https://beta.ruff.rs/) instead of `flake8` and `isort`

### Testing

- [x] Tests do not apply

### Mentions

It might be easier to review by looking at commits individually.

I tried to use the `pyupgrade` fixes from `ruff`, but that broke the tests, and I didn't want to investigate. There are a few other tests that might be interesting to consider in the future: e.g. [`tryceratops`](https://beta.ruff.rs/docs/rules/#tryceratops-try), which checks for how to handle exceptions.


blankdots (commented):
not sure I understand the need for both tox and pre-commit workflows

blankdots:
`pre-commit run --all-files -c .pre-commit-config.yaml` gives
```
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1

swift_browser_ui/request/db.py:48:37: S311 Standard pseudo-random generators are not suitable for cryptographic purposes
swift_browser_ui/request/db.py:51:37: S311 Standard pseudo-random generators are not suitable for cryptographic purposes
swift_browser_ui/request/db.py:54:37: S311 Standard pseudo-random generators are not suitable for cryptographic purposes
swift_browser_ui/sharing/db.py:53:37: S311 Standard pseudo-random generators are not suitable for cryptographic purposes
swift_browser_ui/sharing/db.py:56:37: S311 Standard pseudo-random generators are not suitable for cryptographic purposes
swift_browser_ui/sharing/db.py:59:37: S311 Standard pseudo-random generators are not suitable for cryptographic purposes
swift_browser_ui/ui/middlewares.py:74:11: C901 `error_middleware` is too complex (14 > 10)
Found 7 errors.

```
Maybe we should fix those as well ?

csc-felipe:
> not sure I understand the need for both tox and pre-commit workflows

pre-commit doesn't run pytest. And tox is good for running the same tests with different environment configuration. Also, keeping it there might show issues with the pre-commit configuration, as it's happening now.

csc-felipe:
> `pre-commit run --all-files -c .pre-commit-config.yaml` gives

Did you run `pre-commit autoupdate`? Those checks are in a new version of `ruff`, and I kept the same version from the previous PR so I wouldn't have to fix those now.

blankdots:
> Did you run `pre-commit autoupdate`? Those checks are in a new version of `ruff`, and I kept the same version from the previous PR so I wouldn't have to fix those now.

yes i did :), well i don't see why not update

csc-felipe:
> Weird that mypy has different output when running with pre-commit

If I remove installing `requirements.txt` from tox, mypy doesn't complain. Adding it to pre-commit config doesn't work

blankdots:
> If I remove installing `requirements.txt` from tox, mypy doesn't complain. Adding it to pre-commit config doesn't work

i think that is not the solution, as `mypy` is correct to complain at that line, i think the problem is in pre-commit, I saw some something related to: https://github.com/pre-commit/mirrors-mypy/issues/3

removing `additional_dependencies` and adding language system seems to show the errors, but the additional dependencies need to be installed separately

csc-felipe:
Rebased and fixed the mypy issue. Now it runs from locally installed mypy, and added type deps to `pyproject.toml`

blankdots:
i am ok with the changes, but tests do need to work again.

I remember @csc-jm was working on them, maybe he is interested in this PR

csc-felipe:
Reverted the exception handling to previous code, and used a simpler approach to reduce complexity.

blankdots (approved):

blankdots (approved):

PR #1041, opened 2023-03-24
Fix for selecting and deleting subfolders and files

hannyle:
### Description

- Subfolders and files should be all selectable by checking checkbox
- Subfolders shouldn't be deleted before deleting all files inside it
- Files should be able to select and mass delete

### Related issues
GL 933 & 936 & 974

### Type of change

### Changes Made

- Added folder icon for sub-folders
- Added `prefix` to file's names so the `file` names are matched when its checkbox is selected
- Changed `Display as text` to `Display as path` to make the term more understandable
- Prevented `subfolders` from being deleted when clicking `Delete` button or selecting together with `files`
- Updated Changelog

### Testing
- [x] Needs testing (manual test)

### Mention

- Since we have now `subfolders` and `files` in the same view, some notifications when selecting and deleting them are not clear enough and still not in correct timing/visual order
- The terms and text are still not so clear since we only have `objects` before, now `objects` could mean `subfolders` and `files``
--> These things should be fixed in another PR while waiting for the UI design from @ainoc

### DIFF
--- a/.github/config/.finnishwords.txt
+++ b/.github/config/.finnishwords.txt
@@ -370,6 +370,7 @@ poistettiin
 poistettu
 poistut
 polku
+polkuina
 postamista
 poudan
 profiili
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,7 +120,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix hiding the pagination of data tables
 - Fix shared objects functionality: visibility, deleting, editing tags
 - URL does not strip the path, when that is present. Also small fixes to make it deployable under the same URL.
-- (GL #933) Fix selecting and mass deleting all files
+- (GL #933) Fix for selecting and mass deleting subfolders and files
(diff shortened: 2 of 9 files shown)


hannyle:
Noticed an issue with this, I will make changes

hannyle:
Ready for review

hannyle:
A follow-up PR will be needed, as I wrote in the **Mention** part

.github/config/.finnishwords.txt
emm1R, line 1: This is not a word?

hannyle: You mean `poluna` ? I just google translated  it :D

emm1R: Then it should be `polkuna`.

sampsapenna (approved):
Works nicely, and displaying only the paths of the current folder recursively is indeed more intuitive than the old behavior. Uploads into nested folders still works as well, so I didn't notice any regressions.

LGTM 👍🏼

